Issue #4, opened 2022-07-08
Implement ensemble selection


RondeauG:
### Description

Some ideas:

- Clustering techniques (such as already implemented in `xclim.ensemble`)
- Performance metrics and weighting
- Unicity metrics & weighting
- ECS weighting (or exclusion of models with unrealistic ECS), possibly using already available CSV tables.

Note: For the most part, these features should be implemented in `xclim.ensemble`, but `xscen` should act as a wrapper if required.

PR #35, opened 2022-07-26
Aggregation functions and other minor fixes

RondeauG:
### Pull Request Checklist:
    - This PR is a prerequisite for #4 

#### Main changes ####
* Major API changes from another branch. 
* New file: aggregate.py
  * `aggregate.climatological_mean` computes rolling means over a specified window.
  * `aggregate.deltas` computes deltas.
  * `aggregate.spatial_mean` computes a spatial means using either `.mean()`, `.interp()` or `xesmf.SpatialAverager()`
* `compute_indicators` now supports disjointed time periods.

* Both `shapely` and `geopandas` would be requirements for `xscen`. Are we fine with that?


xscen/__init__.py
```diff
@@ -1,9 +1,21 @@
...
-from types import ModuleType
+
+# Import the submodules
+from . import (
+    aggregate,
+    biasadjust,
+    catalog,
+    checkups,
+    common,
+    config,
```
juliettelavoie, line 11:
```suggestion
    config,
    ensembles,
```

xscen/indicators.py
```diff
@@ -75,16 +76,18 @@ def compute_indicators(
       Can also be only the "stem", if translations and custom indices are implemented.
       Can be the indicator module directly, or a sequence of indicators or a sequence of
       tuples (indicator name, indicator) as returned by `iter_indicators()`.
+    periods : list
+      list of [start, end] of the contiguous periods to be evaluated, in the case of disjointed datasets.
```
juliettelavoie, line 80: Not super clear what you mean here...

I think contiguous periods would mean periods that have the same border (eg. 2005-2010, 2010-2015). I think we want contiNuous periods ( no holes inside the periods).
```suggestion
      list of [start, end] of continuous periods over which to compute the indicators. This is needed when the time axis of ds contains some jumps in time.
```

xscen/indicators.py
```diff
@@ -75,16 +76,18 @@ def compute_indicators(
...
+      If left at None, the dataset will be considered continuous.
```
juliettelavoie, line 81:
```suggestion
      If None, the dataset will be considered continuous.
```

xscen/aggregate.py
```diff
@@ -0,0 +1,443 @@
...
+      Dataset to use for the computation.
+    window: int
+      Number of years to use for the time periods.
+    min_periods: int
+      For the rolling operation, minimum number of years required for a value to be computed.
+      If left at None, it will be deemed the same as 'window'
+    interval: int
+      Interval (in years) at which to provide an output.
+    periods: list
+      list of [start, end] of the contiguous periods to be evaluated, in the case of disjointed datasets.
```
juliettelavoie, line 44: same comment as with compute_indicators

xscen/aggregate.py
```diff
@@ -0,0 +1,443 @@
...
+      If left at None, the dataset will be considered continuous.
+    to_level : str, optional
+      The processing level to assign to the output.
+      If None, the processing level of the inputs is preserved.
+
+    Returns
+    -------
+    xr.Dataset
+      Returns a Dataset with the requested operation applied over time.
```
juliettelavoie, line 53:
```suggestion
      Returns a Dataset of the climatological mean
```

xscen/aggregate.py
```diff
@@ -0,0 +1,443 @@
...
+    # separate 1d time in coords (day, month, and year) to make climatological mean faster
+    ind = pd.MultiIndex.from_arrays(
+        [ds.time.dt.year.values, ds.time.dt.month.values, ds.time.dt.day.values],
+        names=["year", "month", "day"],
+    )
+    ds_unstack = ds.assign(time=ind).unstack("time")
+
+    # Rolling will ignore jumps in time, so we want to raise an exception beforehand
+    if (not all(ds_unstack.year.diff(dim="year", n=1) == 1)) & (periods is None):
+        raise ValueError("Data is not contiguous. Use the 'periods' argument.")
```
juliettelavoie, line 68:
```suggestion
        raise ValueError("Data is not continuous. Use the 'periods' argument.")
```

xscen/aggregate.py
```diff
@@ -0,0 +1,443 @@
...
+    """
+    Computes the spatial mean, using a variety of available methods.
+
+    Parameters
+    ----------
+    ds: xr.Dataset
+      Dataset to use for the computation.
+    method: str
+      'mean' will perform a .mean() over the spatial dimensions of the Dataset.
+      'interp_coord' will (optionally) find the region's centroid, then perform a .interp() over the spatial dimensions of the Dataset.
```
juliettelavoie, line 271: Not sure why we need this method here ? What do we use it for ?
Also, why the "(optionnally)"? Does it do that when coords are not given ?

juliettelavoie (commented):
Cool!
Are you planning on adding a section to GettingStarted ?
I will try a few tests of your functions on my data.

xscen/aggregate.py
RondeauG: Since it is much faster than SpatialAverager and potentially more accurate than `mean` depending on the specifics, I figured that there might be cases where we want to use `interp(method="linear")` with the centroid of a region or shapefile.

Also, the "optionally" is because it's in theory possible to feed the coordinates directly to `interp` via the `kwargs` argument.

RondeauG:
> Cool! Are you planning on adding a section to GettingStarted ? I will try a few tests of your functions on my data.

I was thinking of making a specific Notebook with `aggregate` and `reduce`, focusing on their use for ensemble selection. I was waiting until everything is ready, but I could start working on it if you want to.

juliettelavoie:
I agree that `reduce` should be in its own notebook. But, I feel like `climatological mean` and `deltas` have their place in Getting Started between indicators and ensemble.

xscen/aggregate.py
juliettelavoie:
```suggestion
      'interp_coord' will find the region's centroid (if coordinates are not fed through kwargs), then perform a .interp() over the spatial dimensions of the Dataset.
```

juliettelavoie:
For `compute_deltas`,  the time axis is missing the horizon of reference. Is this on purpose ? It will always be 0 (or 1), but I would be encline to keep it.
1. It is better to have regular axis without holes. 
2. We need it for ccdp.
3. It is a good validation of what the data is.

RondeauG:
> For `compute_deltas`, the time axis is missing the horizon of reference. Is this on purpose ? It will always be 0 (or 1), but I would be encline to keep it.
> 
> 1. It is better to have regular axis without holes.
> 2. We need it for ccdp.
> 3. It is a good validation of what the data is.

It was on purpose since I didn't see the point of keeping it, but I can change that since you have a use case.

RondeauG:
In the code that I copied for `compute_deltas`, the variables are always renamed. For example, `growing_degree_days` becomes `growing_degree_days_delta_1981_2010`. What do we think of that? Should we make it an option?

juliettelavoie:
> In the code that I copied for `compute_deltas`, the variables are always renamed. For example, `growing_degree_days` becomes `growing_degree_days_delta_1981_2010`. What do we think of that? Should we make it an option?

Making it optional is okay with me. The variable was always renamed because it was always in the same dataset has the 30y mean (no delta). But, in xscen, the mean and the deltas are separate.

juliettelavoie:
> In the code that I copied for `compute_deltas`, the variables are always renamed. For example, `growing_degree_days` becomes `growing_degree_days_delta_1981_2010`. What do we think of that? Should we make it an option?

Making it optional is okay with me. The variable was always renamed because it was always in the same dataset has the 30y mean (no delta). But, in xscen, the mean and the deltas are separate.

RondeauG:
This PR should be OK for a new round of reviews. I also updated the notebooks.

aulemahal (approved):
Je n'ai pas testé les fonctions, mais ça me semble beau!

xscen/aggregate.py
```diff
@@ -0,0 +1,455 @@
...
+
+@parse_config
+def climatological_mean(
+    ds: xr.Dataset,
+    *,
+    window: int = None,
+    min_periods: int = None,
+    interval: int = 1,
+    periods: list = None,
+    to_level: str = None,
```
juliettelavoie, line 29:
```suggestion
    to_level: str = 'climatology',
```
Je mettrais un défaut juste pour que la catalogue n'ait pas une entrée vide si la personne ne spécifie pas de level.

xscen/aggregate.py
```diff
@@ -0,0 +1,455 @@
...
+
+
+@parse_config
+def compute_deltas(
+    ds: xr.Dataset,
+    reference_horizon: str,
+    *,
+    kind: Union[str, dict] = "+",
+    rename_variables: bool = True,
+    to_level: str = None,
```
juliettelavoie, line 143:
```suggestion
    to_level: str = 'delta_climatology',
```

juliettelavoie (approved):
looks good!

PR #51, opened 2022-08-23
Adapt to xscen-509

### LINKED ISSUE #13: attributes created by intake-esm contain illegal characters
`intake-esm` creates global attributes based on the catalog. By default, they follow the format: `intake_esm_attr/(x)`. 

However, `/` is an illegal character when, in xarray, `engine="netcdf4"`. We have no control over that character, and so this will need a fix in `intake_esm` itself. In the meantime, a fix is needed in `xscen` too.

aulemahal:
### Pull Request Checklist:
    - This PR fixes #13 
- [x] If a merge request has been made in parallel to this PR in xscen-notebooks, it is merged and the submodules have been updated. [MR#6](https://gitlab.ouranos.ca/scenario/xscen-notebooks/merge_requests/6)

* Use `cat:` to prefix dataset attributes generated from catalogs.
* New `xscen.utils.get_cat_attrs` that outputs a dictionary of those attributes without the prefix.

Yes (but it comes from `intake-esm`)

### Other information:
The intake-esm change also affects how lists are parsed : 
https://github.com/intake/intake-esm/blob/ba4bb59ca1174369ff45a09aed030db0a7aa49d2/intake_esm/source.py#L100-L105

I.e.: every object that is iterable (like lists or tuples) is formatted as a comma-separated string. I think this doesn't break xscen, but might be worth noting. The only list-like attribute we usually use is `variable`.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,6 +2,14 @@
 History
 =======
 
+v0.3.2 (2022-08-23)
+-------------------
+Contributor to this version: Pascal Bourgault (:user:`aulemahal`).
+
+Breaking changes
+^^^^^^^^^^^^^^^^
+* Following a change in intake-esm xscen now uses "cat:" to prefix the dataset attributes extracted from the catalog. All catalog-generated attributes should now be valid when saving to netCDF. (:issue:`13`, :pull:`51`).
+
 v0.3.0 (2022-08-23)
 -------------------
 Contributors to this version: Gabriel Rondeau-Genesse (:user:`RondeauG`), Juliette Lavoie (:user:`juliettelavoie`), Trevor James Smith (:user:`Zeitsperre`) and Pascal Bourgault (:user:`aulemahal`).
--- a/xscen/utils.py
+++ b/xscen/utils.py
@@ -183,6 +183,15 @@ def natural_sort(_list: list):
     return sorted(_list, key=alphanum_key)
 
 
+def get_cat_attrs(ds: xr.Dataset | dict):
+    """Return the catalog-specific attributes from a dataset or dictionary."""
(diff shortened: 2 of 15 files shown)


HISTORY.rst
```diff
@@ -2,6 +2,14 @@
 History
 =======
 
+v0.3.2 (2022-08-23)
```
RondeauG, line 5: This is because it will get its own release, right ?

xscen/utils.py
```diff
@@ -369,13 +378,13 @@ def clean_up(
...
         - starting with a '^' means check if the string starts with the substring.
-        eg. {'global': ['necessary note', '^cat/'], 'tasmax': 'new_name'}
+        eg. {'global': ['necessary note', '^cat:'], 'tasmax': 'new_name'}
     add_attrs: dict
         Dictionary where the keys are the variables and the values are a another dictionary of attributes.
         For global attrs, use the key 'global'.
         eg. {'global': {'title': 'amazing new dataset'}, 'tasmax': {'note': 'important info about tasmax'}}
     change_attr_prefix: str
-        Replace "cat/" in the catalogue global attrs by this new string
+        Replace "cat:" in the catalogue global attrs by this new string
```
RondeauG, line 387:
```suggestion
        Replace "cat:" in the catalog global attrs by this new string
```

RondeauG (approved):
Looks good!

HISTORY.rst
aulemahal: Yep.

juliettelavoie (approved):
ben beau

aulemahal:
@Zeitsperre  J'ai modifié l'action github pour exclure "docs/notebooks".

PR #58, opened 2022-08-30
Fix 57

### LINKED ISSUE #57: default method dans extract_dataset
### Generic Issue

For tasmax, the default `method` in `resample` is max. This makes sense for hourly to daily resampling, but isn't what the user expects for daily to monthly. In that case, the user would have to explicitely call `method='mean'`.

When `resample` is called by `extract_dataset`, there is no way to set the method.

Either `resample` needs to do a better job at setting the default or `method_per_variable` need to be an argument of 'extract_dataset`.

juliettelavoie:
### Pull Request Checklist:
    - This PR fixes #57 

* default `method` for `resample` now looks at the frequency
* `resample_method` added to `extract_dataset` if we want to override the default for some reason

yes, previously resampling tasmax/tasmin would always use method max/min. Now, this is only the case for resampling to 'D'.

### DIFF
--- a/xscen/extract.py
+++ b/xscen/extract.py
@@ -131,7 +131,7 @@ def extract_dataset(
     xr_open_kwargs: dict = None,
     xr_combine_kwargs: dict = None,
     preprocess: Callable = None,
-    resampling_method: Optional[str] = None,
+    resample_method: Optional[str] = None,
 ) -> Union[dict, xr.Dataset]:
     """
     Takes one element of the output of `search_data_catalogs` and returns a dataset,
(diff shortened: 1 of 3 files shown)


xscen/extract.py
```diff
@@ -131,6 +131,7 @@ def extract_dataset(
     xr_open_kwargs: dict = None,
     xr_combine_kwargs: dict = None,
     preprocess: Callable = None,
+    resample_method: Optional[str] = None,
```
RondeauG, line 134: I think this should be a dictionary, since the method might vary per variable.

juliettelavoie: totaly right

xscen/extract.py
```diff
@@ -367,11 +379,16 @@ def resample(
     var_name = da.name
 
     if method is None:
-        if var_name in CV.resampling_methods.dict:
-            method = CV.resampling_methods(var_name)
+        if var_name in CV.resampling_methods.dict["any"]:
```
aulemahal, line 382: Proposition d'inverser les if : 
```
if target_frequency in CV.resampling_methods.dict and var in CV.resampling_methods.dict[target_frequency]:
    method = CV.resampling_methods(target_frequency)[var]
elif var in CV.resampling.methods.dict['any']:
    method = CV.resampling_methods('any')[var]
...
```
Il est possible d'avoir une méthode non-mean pour une variable donnée qui change selon la fréquence, mais qui a une valeur par défaut dans le "any".

RondeauG (approved):

xscen/extract.py
juliettelavoie: d'accord!

aulemahal (approved):
:green_circle: :+1:

PR #67, opened 2022-09-08
Fix unstack

juliettelavoie:
### Pull Request Checklist:
    - This PR fixes #xyz

* ``xs.utils.stack_drop_nans``/ ``xs.utils.unstack_fill_nan``  format the `to_file`/`coords` string to add the domain and the initial shape. 
* The initial shape is added to the lat and lon attrs to be able to retrive it when it is time to unstack.
* Redifining lon and lat to fix bug in xarray (https://github.com/pydata/xarray/issues/6946)

I don't think so. If you pass a string without {domain} or {shape} as before, the formatting will be ignored.

### DIFF
--- a/xscen/utils.py
+++ b/xscen/utils.py
@@ -213,8 +213,9 @@ def unstack_fill_nan(
 
     if not isinstance(coords, (list, tuple)) and coords is not None:
         if isinstance(coords, (str, os.PathLike)):
-            original_shape = ds.lat.attrs["original_shape"]
-            coords = coords.format(domain=ds.attrs["cat:domain"], shape=original_shape)
+            original_shape = ds.lat.attrs.get("original_shape", "unknown_shape")
+            domain = ds.attrs.get("cat:domain", "unknown_domain")
+            coords = coords.format(domain=domain, shape=original_shape)
             logger.info(f"Dataset unstacked using {coords}.")
             coords = xr.open_dataset(coords)
         out = out.reindex(**coords.coords)
--- a/xscen/utils.py
+++ b/xscen/utils.py
@@ -247,6 +247,7 @@ def get_cat_attrs(ds: Union[xr.Dataset, dict]):
     return {k[4:]: v for k, v in attrs.items() if k.startswith("cat:")}
 
 
+@parse_config
 def maybe_unstack(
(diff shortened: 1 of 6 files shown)


xscen/utils.py
```diff
@@ -106,13 +116,39 @@ def stack_drop_nans(
...
+        .where(mask_1d, drop=True)
+        .reset_index(new_dim, drop=True)
+    )
     for dim in mask.dims:
         out[dim].attrs.update(ds[dim].attrs)
 
     if to_file is not None:
+        # set default path to store the information necessary to unstack
+        # the name includes the domain and the original shape to uniquely identify the dataset
+        domain = ds.attrs.get("cat:domain", "unknown_domain")
```
aulemahal, line 134: Je suggère pas de '_'
```suggestion
        domain = ds.attrs.get("cat:domain", "unknown")
```

xscen/utils.py
```diff
@@ -106,13 +116,39 @@ def stack_drop_nans(
     --------
     unstack_fill_nan : The inverse operation.
     """
+
+    original_shape = f"{len(ds.lon)}x{len(ds.lat)}"
```
aulemahal, line 120: Je sais pas à quel point c'était utile, mais l'ancienne version de cette fonction était indépendante des noms de dimensions. On utilise `mask.dims`.
Est-ce que ça serait à garder?
Si oui, je propose:
```suggestion
    original_shape = "x".join(map(str, mask.shape))
```
et il faudrait rendre le reste du code générique aux dims de `mask`...

Et si non, il faudrait juste le dire dans la fonction qu'on s'attend à `lat` et `lon`.

aulemahal (commented):
Ça me semble une bonne solution!
Petit questionnement sur la généricité (?) à conserver?

xscen/utils.py
juliettelavoie: ah ben oui ça serait bien de garder cette flexibilité là. Je vais faire les changements.

juliettelavoie: The changes I have added will fail if one of the dimensions of ds_stack have an unrelated attrs also named `original_shape`, but that feels very unlikely...

aulemahal (approved):

PR #68, opened 2022-09-08
Bunch of fixes

### LINKED ISSUE #64: Pass in-memory catalog to `search_data_catalogs`
### Addressing a Problem?

The current `search_data_catalog` re-opens the CSV files of the passed DataCatalog(s).
See: https://github.com/Ouranosinc/xscen/blob/6285df36e2fa5297d88871adee37ee6e9d26bdad/xscen/extract.py#L549-L560

The problem arises when on has made in-memory changes to the DataCatalog, which are not yet included in the csv. In the case of a ProjectCatalog, one can write the changes to disk, but even then, I find it non-intuitive.

### Potential Solution

I think I would make sense to use the dataframe(s) of the passed DataCatalog(s) instead of going back to the csvs.

### Additional context

I believe the current situation arises from the need to have a DataCatalog with a "registry" when `allow_conversion` is True. This might mean we would have to implement something to copy a DataCatalog while adding a registry.

### Contribution

aulemahal:
### Pull Request Checklist:
    - This PR fixes #64.

* Add a missing dependencies to the env. (pyarrow for faster string handling in catalogs)
* New `unstack_dates` function to "extract" seasons or months from a timeseries.
* Better `spatial_mean` for cases using xESMF and a shapefile with multiple polygons.
* Yet more changes to `parse_directory`:
    * Better parallelization by merging the finding and name-parsing step in the same dask tree.  
    * Allow cvs for the variable columns
    * Fix parsing the variable names from datasets
    * Sort the variables in the tuples (for a more consistent output)
* `extract_dataset`
    * Option to ensure the time coordinate matches the expected freq. Ex: monthly values given on the 15th day are moved to the 1st, as expected when asking for "MS".
    * Lower the verbosity of some logging calls
*  Allow passing `compute=False` to `save_to_zarr`.
*  Allow passing `skipna` to the regridder kwargs.
*  Regrid : Do not fail if a grid_mapping attribute mentions a variable that doesn't exist.
*  In fact, do not fail for any grid mapping problem, simply warn about it.
*  Default email sent to the local user.
*  DataCatalogs passed to `search_data_catalogs` are used as is, instead of re-reading their csv file (that might be out of date).

I do not think so.

### Other information:
@tlogan2000  : See `xscen.aggregate.unstack_dates` and tell me if you like it!

The email thing allows to avoid mentioning an email adress in the configl. On normal unix system, you can create the "~/.forward" file with your real email address  in it. Then, when mail is sent to "$USER@$HOSTNAME" it gets redirected to you.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -18,7 +18,7 @@ New features and enhancements
     - Allow cvs for the variable columns
     - Fix parsing the variable names from datasets
     - Sort the variables in the tuples (for a more consistent output)
-* In extract_dataset, add option ``ensure_correct_time`` to ensure the time coordinate matches the expected freq. Ex: monthly values given on the 15th day are moved to the 1st, as expected when asking for "MS".
+* In extract_dataset, add option ``ensure_correct_time`` to ensure the time coordinate matches the expected freq. Ex: monthly values given on the 15th day are moved to the 1st, as expected when asking for "MS". (:issue: `53`).
 * In regrid_dataset: (:pull:`68`).
     * Allow passing skipna to the regridder kwargs.
     * Do not fail for any grid mapping problem, includin if a grid_mapping attribute mentions a variable that doesn't exist.
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -16,7 +16,7 @@ Breaking changes
 
 Internal changes
 ^^^^^^^^^^^^^^^^
-* Small bugfixes in aggregate.py (:pull:`55`).
+* Small bugfixes in aggregate.py (:pull:`55`, :pull:`56`).
 
 v0.3.0 (2022-08-23)
 -------------------
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -11,6 +11,7 @@ New features and enhancements
 * New functions ``diagnostics.properties_and_measures``, ``diagnostics.measures_heatmap`` and ``diagnostics.measures_improvement``. (:issue:`5`, :pull:`54`)
 * Add argument `resample_methods` to `xs.extract.resample`. (:issue:`57`, :pull:`57`)
 * Added a ReadTheDocs configuration to expose public documentation. (:issue:`65`, :pull:`66`).
+* ``xs.utils.stack_drop_nans``/ ``xs.utils.unstack_fill_nan`` will now format the `to_file`/`coords` string to add the domain and the shape. (:issue:`59`, :pull:`67`)
 * New unstack_dates function to "extract" seasons or months from a timeseries. (:pull:`68`).
 * Better spatial_mean for cases using xESMF and a shapefile with multiple polygons. (:pull:`68`).
 * Yet more changes to parse_directory: (:pull:`68`).
@@ -23,7 +24,6 @@ New features and enhancements
     * Allow passing skipna to the regridder kwargs.
     * Do not fail for any grid mapping problem, includin if a grid_mapping attribute mentions a variable that doesn't exist.
 * Default email sent to the local user. (:pull:`68`).
-
 
 Breaking changes
 ^^^^^^^^^^^^^^^^
@@ -38,6 +38,7 @@ Internal changes
 ^^^^^^^^^^^^^^^^
 * Small bugfixes in aggregate.py (:pull:`55`, :pull:`56`).
 * Default method of `xs.extract.resample` now depends on frequency. (:issue:`57`, :pull:`58`).
+* Bugfix for `_restrict_by_resolution` with CMIP6 datasets (:pull:`71`).
 
 v0.3.0 (2022-08-23)
 -------------------
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -9,6 +9,7 @@ Contributors to this version: Gabriel Rondeau-Genesse (:user:`RondeauG`), Juliet
 New features and enhancements
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 * New functions ``diagnostics.properties_and_measures``, ``diagnostics.measures_heatmap`` and ``diagnostics.measures_improvement``. (:issue:`5`, :pull:`53`)
+* Add argument `resample_methods` to `xs.extract.resample`. (:issue:`57`, :pull:`57`)
 
 Breaking changes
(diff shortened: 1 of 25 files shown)


juliettelavoie (approved):
Just comments on the docs.
J'avoue que je suis un peu perdu dans ce que tu fais dans catalog.py, mais le reste m'a l'air bon!

xscen/aggregate.py
```diff
@@ -272,6 +273,7 @@ def spatial_mean(
       If method=='xesmf', the bounding box or shapefile is given to SpatialAverager.
     kwargs: dict
       Arguments to send to either mean(), interp() or SpatialAverager().
+      In the latter case, one can give `skipna` here, to be passed to the averager call itself.
```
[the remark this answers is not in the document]
aulemahal: Oh, "latter case" comme dans "SpatialAverager". Je vais mettre ça plus clair.

aulemahal:
Ouain, je vais essayer de mettre un peu plus de commentaires. L'idée c'est de n'appeller  `dask.compute` qu'une seule fois pour optimiser le parallélisme.  Avant, je voyais que la recherche dans un des sous-dossiers prenaient beaucoup plus de temps que les autres et on devait attendre avant de passer à la prochaine étape. Là, l'analyse des noms de fichiers peut commencer dès qu'un coeur se libère.

HISTORY.rst
```diff
@@ -11,11 +11,29 @@ New features and enhancements
...
 * Add argument `resample_methods` to `xs.extract.resample`. (:issue:`57`, :pull:`57`)
 * Added a ReadTheDocs configuration to expose public documentation. (:issue:`65`, :pull:`66`).
+* New unstack_dates function to "extract" seasons or months from a timeseries. (:pull:`68`).
+* Better spatial_mean for cases using xESMF and a shapefile with multiple polygons. (:pull:`68`).
+* Yet more changes to parse_directory: (:pull:`68`).
+    - Better parallelization by merging the finding and name-parsing step in the same dask tree.
+    - Allow cvs for the variable columns
+    - Fix parsing the variable names from datasets
+    - Sort the variables in the tuples (for a more consistent output)
+* In extract_dataset, add option ``ensure_correct_time`` to ensure the time coordinate matches the expected freq. Ex: monthly values given on the 15th day are moved to the 1st, as expected when asking for "MS".
```
juliettelavoie, line 21:
```suggestion
* In extract_dataset, add option ``ensure_correct_time`` to ensure the time coordinate matches the expected freq. Ex: monthly values given on the 15th day are moved to the 1st, as expected when asking for "MS". (:issue: `53`).
```

xscen/aggregate.py
```diff
@@ -457,3 +465,84 @@ def spatial_mean(
         ds_agg.attrs["cat:processing_level"] = to_level
 
     return ds_agg
+
+
+@parse_config
+def unstack_dates(
```
RondeauG, line 471: Why is this in `aggregate`? Shouldn't it be in `utils`?

xscen/io.py
```diff
@@ -386,7 +386,7 @@ def save_to_zarr(
 
     Returns
     -------
-    None
+    dask.delayed object if compute=False, None otherwise.
```
RondeauG, line 389: Just curious, what's the use case of doing a `save_to_zarr` with `compute=False`?

RondeauG (approved):
There are a lot of small changes so I might have missed something, but it looks good to me.

xscen/io.py
aulemahal: Saving to datasets at the same time! I had a use for it with the diagnostics : I don't want to redo the same computation twice, so I want the saving of the properties and of the measures to be done at the same time.

xscen/aggregate.py
aulemahal: Ouain, it could move. It's there because it's the kind of thing that is done that the same moment as aggregations.

xscen/extract.py
```diff
@@ -315,6 +321,14 @@ def extract_dataset(
                         catalog[key].df["xrfreq"].iloc[0]
                         == variables_and_freqs[var_name]
                     ):
+                        if ensure_correct_time:
+                            counts = da.time.resample(time=xrfreq).count()
+                            if any(counts > 1):
+                                raise ValueError(
+                                    "Dataset is labelled as having a sampling frequency of "
+                                    f"{xrfreq}, but some periods have more than one data point."
+                                )
```
juliettelavoie, line 330:
```suggestion
                            if any(counts > 1):
                                raise ValueError(
                                    "Dataset is labelled as having a sampling frequency of "
                                    f"{xrfreq}, but some periods have more than one data point."
                                )
                            if any(counts.isnull()):
                                warnings.warn('The resampling count contains nans. There might be some missing data.')
```
If there is a whole in the data, this might be a way to catch it. Related to issue #70. I am working on a fix for the check in `search`, but I think we also need to implement a check in `extract`.

If we add this, we also need to import warnings at the top. I put a warning not an error because you might want to live with the hole. I'm not sure if this should be in the logger instead?

juliettelavoie: Just realised that if there is a whole, this will fail on `da["time"] = counts.time`, because the shape won't be that same. Maybe, we need to raise an error instead of a warning.
```suggestion
                            if any(counts > 1):
                                raise ValueError(
                                    "Dataset is labelled as having a sampling frequency of "
                                    f"{xrfreq}, but some periods have more than one data point."
                                )
                            if any(counts.isnull()):
                                raise ValueError('The resampling count contains nans. There might be some missing data.')
```

aulemahal: Indeed! We need this check. Otherwise we would haul all whole holes!

juliettelavoie: ishhh j'ai de la misère avec mes homophones!

aulemahal: ses pas grave

aulemahal:
Je pense que cette PR est sans impact réel sur les notebooks, mais il vous faudra supprimer les *.zarr dans `docs/notebooks/samples` après le merge. Un des datasets de `getting_started` est daily à midi et `xs.save_to_zarr(..., mode='o')` ne va overwrite QUE la variable. PAS les coords. Ça donnait un `time` brisé, en tout cas pour moi.

PR #75, opened 2022-09-13
Accelerate catalog parsing

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

The default "read csv kwargs" not contain a special function for parsing `date_start` and  `date_end`. Instead of directly parsing from string to period (and element by element), this function parses the whole array to a `DatetimeIndex` which it then converts to a `PeriodIndex`. If an `OutOfBoundsDatetime` error is raised, it switches back to parsing the dates with `PeriodIndex` directly.

If the direct-period version is used, a warning is issued, so the user knows they have time to go and get an orange juice.

No.

### Other information:
Parsing strings with `Period` and `PeriodIndex` is ridiculously slow. See for yourselves:

- The complete "simulations" catalog, 95 000 entries, with some dates after 2262-04-12, thus it must use the direct-period method (similar to before this PR): 50 s is spent converting the dates.
- The MRCC5 C series catalog, 950 000 entries (10x more), all dates within the `datetime64[ns]` range : 0.7 s.

I think I may raise the issue to pandas. But for now, we have a way around.

EDIT: Oops, I forgot that, but this also adds an error catch and logging call for some errors I got in `name_parser`.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -24,6 +24,7 @@ New features and enhancements
     * Allow passing skipna to the regridder kwargs.
     * Do not fail for any grid mapping problem, includin if a grid_mapping attribute mentions a variable that doesn't exist.
 * Default email sent to the local user. (:pull:`68`).
+* Special accelerated pathway for parsing catalogs with all dates within the datetime64[ns] range (:pull:`75`).
 
 Breaking changes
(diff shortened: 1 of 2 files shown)


juliettelavoie (approved):
nice!

RondeauG (approved):

PR #77, opened 2022-09-19
simpler way to create a new catalog

### LINKED ISSUE #11: Simplify ProjectCatalog.create()
### Description

Currently, if I want to rerun a workflow without overwriting the ProjectCatalog, I need something like:

```
if os.path.isfile(catalog_path):
   pcat = ProjectCatalog(catalog_path)
else:
   pcat = ProjectCatalog.create(catalog_path, project=project, overwrite=True)
```

I think we could instead have a one-liner like:

`pcat = ProjectCatalog(catalog_path, project=project, create_if_required=True, overwrite=False)`

RondeauG:
### Pull Request Checklist:
    - This PR fixes #11

* Adds a `create` argument to `ProjectCatalog()`:
  * `ProjectCatalog("path.json", create=True, project=project, overwrite=False)` (default) will only create a new catalog if one doesn't already exist. If it exists, it will drop the creation-specific kwargs and open the JSON normally.
  *  `ProjectCatalog("path.json", create=True, project=project, overwrite=True)` will create a new catalog, overwriting if necessary.
  * Any version without `create` will work as previously and crash if the file doesn't exist.

No.

### DIFF
--- a/xscen/catalog.py
+++ b/xscen/catalog.py
@@ -410,7 +410,38 @@ class ProjectCatalog(DataCatalog):
 
         return cls(str(meta_path))
 
-    def __init__(self, df, *args, create=False, overwrite=None, project=None, **kwargs):
+    def __init__(
+        self,
+        df: Union[str, dict],
+        *args,
+        create: bool = False,
+        overwrite: bool = None,
+        project: dict = None,
+        **kwargs,
+    ):
+        """
+        Open or create a project catalog.
+
+        Parameters
+        ----------
+        df : PathLike, dict
+          If string, this must be a path or URL to a catalog JSON file.
+          If dict, this must be a dict representation of an ESM catalog.  See the notes below.
+        create: bool
+          If True, and if 'df' is a string, this will create an empty ProjectCatalog if none already exists.
+        project : dict-like
+          Metadata to create the catalog, if required.
+        overwrite : bool
+          If this and 'create' are True, this will overwrite any existing JSON and CSV file with an empty catalog.
+
+        Notes
+        ----------
+        The dictionary in 'df' must have two keys: ‘esmcat’ and ‘df’.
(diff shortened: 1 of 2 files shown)


xscen/catalog.py
```diff
@@ -411,6 +411,14 @@ def create(
...
 
     def __init__(self, df, *args, **kwargs):
+        if "create" in kwargs:
+            kwargs.pop("create")
+            if isinstance(df, (str, Path)) and (
+                not os.path.isfile(Path(df)) or kwargs.get("overwrite", None) is True
+            ):
+                self.create(df, **kwargs)
+        kwargs.pop("project", None)
+        kwargs.pop("overwrite", None)
```
aulemahal, line 429:
```suggestion
    def __init__(self, df, *args, create=False, overwrite=None, project=None, **kwargs):
        if create:
            if isinstance(df, (str, Path)) and (not os.path.isfile(Path(df)) or overwrite):
                self.create(df, project=project, overwrite=overwrite)
```
If kwargs are not passed directly to a parent, I find it makes more sense to have them explicitly in the signature.

But in fact I am wondering if your signatures make sense at all here. When I dig, I can't see what `*args` could be and the only meaningful `kwarg` are `read_csv_kwargs` and `registry`. Why not list them directly? I can push something if you agree.

aulemahal: (outdated comment, but I think it addresses your last changes too)

RondeauG: I'm honestly not sure either... I think we might just have used `*args, **kwargs` as a default way to support whatever changes could occur in `intake-esm`?

aulemahal: Indeed, it's a good idea to be ready for future changes... In that case, then I think this is satisfactory!

aulemahal (approved):

PR #88, opened 2022-09-27
add cf names to conversions

### LINKED ISSUE #86: Use variable names for conversions, instead of the xclim indicator
### Generic Issue

* xscen version: 0.3.17

### Description
If I have a catalog with `tas` and `pr`, I would expect `prsn` to be found with `search_data_catalogs(allow_conversion=True, ...)`, but it's not. Instead, I need to search for `snowfall_approximation`. Once computed, the variable also appears as `snowfall_approximation` instead of `prsn` in the dataset`

Also, do we really want `allow_conversion=True` to be the default? I feel like this is something that should be a conscious choice.

RondeauG:
### Pull Request Checklist:
- [ ] `pre-commit` hooks are installed/active in my local clone (`$ pre-commit install`)
    - This PR fixes #86

* Where it was missing, added CF names to conversions.

No.

### DIFF
--- a/xscen/xclim_modules/conversions.yml
+++ b/xscen/xclim_modules/conversions.yml
@@ -21,26 +21,36 @@ indicators:
     parameters:
       method: tetens30
       ice_thresh: 0 degC
+    cf_attrs:
+      - var_name: hurs
   relative_humidity:
     base: hurs
     parameters:
       method: tetens30
       ice_thresh: 0 degC
+    cf_attrs:
+      - var_name: hurs
   specific_humidity:
     base: huss
     parameters:
       method: tetens30
       ice_thresh: 0 degC
+    cf_attrs:
+      - var_name: huss
   snowfall_approximation:
     base: prsn
     parameters:
       method: binary
       thresh: 0 degC
+    cf_attrs:
+      - var_name: prsn
   rain_approximation:
     base: prlp
     parameters:
       method: binary
       thresh: 0 degC
+    cf_attrs:
+      - var_name: prlp
   precipitation:
     compute: precipitation
     cf_attrs:
@@ -58,6 +68,8 @@ indicators:
         kind: 0
       tasmax:
         kind: 0
+    cf_attrs:
+      - var_name: evspsblpot
   dtr:
     src_freq: D
     compute: dtr


aulemahal (approved):
I'm not sure about this, but I think this is actually a regression/breaking change coming from xclim! I'm pretty sure the initial version of xscen with an older xclim was using those variable names.

This fixes it! Thanks.

HISTORY.rst
```diff
@@ -54,6 +54,8 @@ Internal changes
 * Use numpy max and min in `properties_and_measures` (:pull:`76`).
 * Cast catalog date_start and date_end to "%4Y-%m-%d %H:00" when writing to disk. (:issue:`83`, :pull:`79`)
 * Skip test of coverage on the sum if the list of select files is empty. (:pull:`79`)
+* Added missing cf names in conversions.yml (:issue:`86`, :pull:`88`)
```
aulemahal, line 57:
```suggestion
* Added missing CMIP variable names in conversions.yml (:issue:`86`, :pull:`88`)
```
The variable codes are usually taken from CMIP6 in xclim. The CF conventions do not define those codes, only the standard names and "canonical" units.

RondeauG:
Do you think that there's an issue with `relative_humidity` and `relative_humidity_from_dewpoint` having the same variable name? I don't really have the data to test it out...

aulemahal:
I don't believe so, but I can test it.

aulemahal:
Woups yes it does, the registry is a dictionary where keys are the variable names. So only one of the `hurs` function gets registered, the last one in the yaml.

juliettelavoie:
Donc, dans `variable_and_freqs`, on veut que les utilisateurs mettent ce qui est dans  [cf_attrs][var_name], right?
Si une personne demande `hurs`, lequel va être utilisé ?

RondeauG:
> Woups yes it does, the registry is a dictionary where keys are the variable names. So only one of the `hurs` function gets registered, the last one in the yaml.

Welp. In that case, I suggest keeping only one (`relative_humidity_from_dewpoint` ?) and exposing `conversions.yml` as an argument, so that people can provide their own version as needed. I think it's a decent compromise?

xscen/extract.py
```diff
@@ -506,8 +506,9 @@ def search_data_catalogs(
     match_hist_and_fut: bool = False,
     periods: list = None,
     id_columns: Optional[List[str]] = None,
-    allow_resampling: bool = True,
-    allow_conversion: bool = True,
+    allow_resampling: bool = False,
+    allow_conversion: bool = False,
+    conversion_yaml: str = None,
```
aulemahal, line 511: What do you think about accepting `Union[bool, str]` for  `allow_conversion`? So if it is True, it uses the default file and if it is a str/path, it uses that file? 
It's ok if you think this is confusing.

aulemahal (approved):
Small idea, but this is already good.

xscen/extract.py
RondeauG: Yeah, I fear that this might end up being more confusing than is worth :/

xscen/extract.py
```diff
@@ -535,8 +536,11 @@ def search_data_catalogs(
       If True (default), variables with a higher time resolution than requested are considered.
     allow_conversion: bool
       If True (default) and if the requested variable cannot be found, intermediate variables are
-      searched given that there exists a converting function in the "derived variable registry"
-      defined by "xclim_modules/conversions.yml".
+      searched given that there exists a converting function in the "derived variable registry".
+    conversion_yaml: str
+      Path to a YAML file that defines the possible conversions (used alongside 'allow_conversion=True').
```
RondeauG, line 541:
```suggestion
      Path to a YAML file that defines the possible conversions (used alongside 'allow_conversion'=True).
```

HISTORY.rst
```diff
@@ -54,6 +54,8 @@ Internal changes
 * Use numpy max and min in `properties_and_measures` (:pull:`76`).
 * Cast catalog date_start and date_end to "%4Y-%m-%d %H:00" when writing to disk. (:issue:`83`, :pull:`79`)
 * Skip test of coverage on the sum if the list of select files is empty. (:pull:`79`)
+* Added missing CMIP variable names in conversions.yml (:issue:`86`, :pull:`88`)
```
RondeauG, line 57:
```suggestion
* Added missing CMIP variable names in conversions.yml and added the ability to provide a custom file instead (:issue:`86`, :pull:`88`)
```

PR #92, opened 2022-09-28
Fixes to search_data_catalogs

### LINKED ISSUE #42: search_data_catalog crashes if the result is empty
### Setup Information

- xscen version: 0.2.3

### Description

If `search_data_catalog` fails to find valid entries, the code crashes instead of returning a warning and/or an empty dictionary.

### Steps To Reproduce

```
 cat = xs.extraction.search_data_catalogs(DC, variables_and_freqs={"tas": "D"}, other_search_criteria={"source": "make_it_crash"})
```

Will result in:
```
Traceback (most recent call last):
  File ".../pycharm-community-2017.3.4/helpers/pydev/_pydevd_bundle/pydevd_exec2.py", line 3, in Exec
    exec(exp, global_vars, local_vars)
  File "<string>", line 1, in <module>
  File ".../xscen/config.py", line 163, in _wrapper
    return func(*args, **kwargs)
  File ".../xscen/extraction.py", line 149, in search_data_catalogs
    for (sim_id,), scat in catalog.iter_unique("id"):
  File ".../xscen/catalog.py", line 198, in iter_unique
    for values in itertools.product(*map(self.unique, columns)):
  File ".../xscen/catalog.py", line 186, in unique
    raise ValueError("Catalog is empty.")
ValueError: Catalog is empty.
```

### Additional context

_No response_

### Contribution

### LINKED ISSUE #81: Manage fixed fields in both historical and future periods
### Setup Information

- xscen version: 0.3.16

### Description

`search_data_catalogs(..., match_hist_and_fut=True)` will sometimes create doubles of fixed fields, when they exist for both the historical and future time periods. This subsequently makes the call to `extract_dataset()` fail, because it attempts to `concat` them alongside the `time` dimensions, but fails to find it. 

### Steps To Reproduce

_No response_

### Additional context

This is probably due to this:
```
    "aggregation_control": {
        "variable_column_name": "variable",
        "groupby_attrs": ["id", "domain", "processing_level", "xrfreq"],
        "aggregations": [
            {
                "type": "join_existing",
                "attribute_name": "date_start",
                "options": {"dim": "time"},
            },
            {"type": "union", "attribute_name": "variable"},
        ],
```

### Contribution

RondeauG:
### Pull Request Checklist:
    - This PR fixes #42
    - This PR fixes #81

* `search_data_catalogs` no longer crashes if it finds nothing. It now returns an empty dictionary and a warning.
* `_dispatch_historical_to_future` no longer duplicates fixed fields if they exist in both periods.
* If a fixed field is missing from the search, it is looked for in other experiments or members.


RondeauG:
Question: 
I have 3 members for the same model: 'r1i1p1f1_NorESM2-MM_gn_NCC', 'r2i1p1f1_NorESM2-MM_gn_NCC', and  'r3i1p1f1_NorESM2-MM_gn_NCC'. For some reason, `r1, r2, and r3` all have `orog` and `areacella`, but only `r2` has `sftlf`.

Should we provide an option to broadcast fixed fields to other members or is that too dangerous?

Zeitsperre:
> Should we provide an option to broadcast fixed fields to other members or is that too dangerous?

I can say that for CORDEX-related datasets, this is how the fixed field data is presented between experiment runs. If you feel it's dangerous, you could add warnings when this is being assumed.

xscen/extract.py
```diff
@@ -628,6 +624,10 @@ def search_data_catalogs(
             f"Removing {len(ex.df)} assets based on exclusion dict : {exclusions}."
         )
 
+    if match_hist_and_fut:
```
aulemahal, line 627: Wait. How to you select an experiment with this new order ?
If I want "rcp45", I would I put it in `other_search_criteria`, does this mean I now have to say : `experiment=['historical', 'rcp45']` ?

aulemahal (commented):
I have a few changes I want to make to `_dispatch_historical_future` from an issue encountered by Chris. I could do them here, but it can also wait for another PR.

xscen/extract.py
RondeauG: Right... I didn't use `experiment` in my test, but it would indeed be an issue. Do you have an idea on a possible compromise? This order is a lot faster, but I'm not against reverting this change if we need to.

RondeauG: Alright, let's forget it for now, then.

juliettelavoie (approved):

PR #104, opened 2022-10-11
Fix 102

### LINKED ISSUE #102: converting 360_day calendar create temperature inversion
### Generic Issue

* xscen version: xs.__version__
* Python version: 3.9.10
* Operating System: mac

In `clean_up`, when we convert a 360_day to a standard calendar with the default xscen parameters, we create temperature inversions.

In xscen, the default `align_on` is set to `random` if the input ds is a 360_day calendar. However, it looks like the random dates that are choosen to expand the calendar to standard are not the same for all the variables in the dataset. Eg. `xclim.convert_calendar` might insert a nan on January 23rd for tasmin, but insert a nan on January 31st for tasmax. 

Next, xscen will interpolate on those nan. This has the effect of shifting the timeseries. In some occasions, shifting tasmax but not tasmin with create a temperature inversion (tasmin>tasmax). 

The top lines are tasmax, the bottom lines are tasmin.

image

Potential solutions:
1. Set default align_on year. The down side is that the the same date will always be interpolated. 
2. Change xclim to make random the same for tasmax et tasmin ?

juliettelavoie:
### Pull Request Checklist:
    - This PR fixes #102 

* Instead of re-converting the calendar for each variable individually, use the conversion that was done on the whole dataset pour variable that interpolate over missing dates.
* fixes #102 

not sure? The conversion will be done at the same time for many variables instead of one at the time. Hence, the dates that are interpolated over will be the same for all variables, instead of being different for all variables. So, you can't reproduce the old result. But it was already impossible to reproduce it exactly because the dates are random.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,6 +2,26 @@
 History
 =======
 
+v0.5.0 (unreleased)
+-------------------
+Contributors to this version: Gabriel Rondeau-Genesse (:user:`RondeauG`), Juliette Lavoie (:user:`juliettelavoie`), Trevor James Smith (:user:`Zeitsperre`) and Pascal Bourgault (:user:`aulemahal`).
+
+New features and enhancements
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+
+Breaking changes
+^^^^^^^^^^^^^^^^
+
+Bug fixes
+^^^^^^^^^
+* ``clean_up`` now converts the calendar of variables that use "interpolate" in "missinge_by_var" at the same time.
+Hence, when it is a conversion from a 360_day calendar, the random date are the same of the these variables. (:issue:`102`, :pull:`104`)
+
+Internal changes
+^^^^^^^^^^^^^^^^
+
+
 v0.4.0 (2022-09-28)
 -------------------
 Contributors to this version: Gabriel Rondeau-Genesse (:user:`RondeauG`), Juliette Lavoie (:user:`juliettelavoie`), Trevor James Smith (:user:`Zeitsperre`) and Pascal Bourgault (:user:`aulemahal`).
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -9,7 +9,6 @@ Contributors to this version: Gabriel Rondeau-Genesse (:user:`RondeauG`), Juliet
 New features and enhancements
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-
 Breaking changes
 ^^^^^^^^^^^^^^^^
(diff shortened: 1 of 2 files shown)


HISTORY.rst
```diff
@@ -2,6 +2,24 @@
...
+New features and enhancements
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Breaking changes
+^^^^^^^^^^^^^^^^
+
+Bug fixes
+^^^^^^^^^
+* ``clean_up`` now converts the calendar of variables that use "interpolate" in "missinge_by_var" at the same time.
+Hence, when it is a conversion from a 360_day calendar, the random date are the same of the these variables. (:issue:`102`, :pull:`104`)
```
aulemahal, line 18:
```suggestion
Hence, when it is a conversion from a 360_day calendar, the random dates are the same for all of the these variables. (:issue:`102`, :pull:`104`)
```

aulemahal (approved):
This looks good for the interpolate version!
I see you didn't change the other version, but I think the problem would still occur there, no? However, I can see how this alternative would be less common...

juliettelavoie:
Well, it doesn't work to just use the conversion done on l.548 for the other version because we want the missing argument to be something else than nan.

I guess I could use it and replace the nan by the missing value wanted, but I thought there might be a risk of  filling nans that are not there because the calendar conversion...

I couldn't think of an example where we needed the variables with a fix missing value to be the same dates for all variables. It might even be better that the missing pr and missing temperature dates are not the same ?

aulemahal:
Indeed, filling nans is dangerous. But if the temperature inversion you noticed is a clear-cut physical inconsistency, the one day lag between temperature and pr is also one, even if more subtle. It could yield completely wrong results for indicators like "cold_and_wet_days".

aulemahal:
I guess that for the special case where `missing_by_var` has other values than `interpolate` and the src (or target) calendar is 360_day, you could:
- Convert the dataset, with missing set to some flag value. 
- When interpolating, replace those flags to nan first.
- Replace the flags with the missing_by_var choice for the other case.

I'm not sure what a good flag value would be. It needs to be a float, but  unique. Maybe something like -9999 , which is highly improbable for all common variables we use?
I am aware this change is not elegant... Would you prefer one of the xclim solution I suggested? Then we could leave this PR the current way and update the code once xclim is updated.

juliettelavoie:
ok I see. I can put a unique weird value in the convert on the whole dataset, then use where to fill the missing dates.

juliettelavoie:
I think I'm okay with using the flag.

PR #106, opened 2022-10-17
Minor fixes following ESPO-R5's diagnostics

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

* New "timeout_cleanup" option for ``save_to_zarr``, which removes variables that were in the process of being written when receiving a ``TimeoutException``. (:pull:`106`).
* New ``scripting.skippable`` context, allowing the use of CTRL-C to skip code sections. (:pull:`106`)
* * ``properties_and_measures`` no longer casts month coordinates to string (:pull:`106`).
* Fixed `parse_config`. There is a very special option that logs what args were parsed from the config. The log calls have been fixed,
* Added a bit more information in the "insufficient coverage" log warning.

No.

### Other information:
These changes are the ones I used to help me run my "properties and diagnostics" script for ESPO-R5. Forget the branch name it is meaningless.

### DIFF
--- a/xscen/diagnostics.py
+++ b/xscen/diagnostics.py
@@ -244,8 +244,6 @@ def properties_and_measures(
         # to be able to save in zarr, convert object to string
         if "season" in ds1:
             ds1["season"] = ds1.season.astype("str")
-        if "month" in ds1:
-            ds1["month"] = ds1.month.astype("str")
 
     prop.attrs["cat:processing_level"] = to_level_prop
     meas.attrs["cat:processing_level"] = to_level_meas
--- a/xscen/io.py
+++ b/xscen/io.py
@@ -13,6 +13,7 @@ from rechunker import rechunk as _rechunk
 from xclim.core.calendar import get_calendar
 
 from .config import parse_config
+from .scripting import TimeoutException, timeout
 from .utils import translate_time_chunk
 
 logger = logging.getLogger(__name__)
@@ -358,6 +359,7 @@ def save_to_zarr(
     encoding: dict = None,
     mode: str = "f",
     itervar: bool = False,
+    timeout_cleanup: bool = False,
 ) -> None:
     """
     Saves a Dataset to Zarr, rechunking if requested.
@@ -385,6 +387,10 @@ def save_to_zarr(
     itervar : bool
       If True, (data) variables are written one at a time, appending to the zarr.
       If False, this function computes, no matter what was passed to kwargs.
+    timeout_cleanup : bool
+      If True and a :py:class:`xscen.scripting.TimeoutException` is raised during the writing,
+      the variable being written is removed from the dataset as it is incomplete.
+      This does nothing if `compute` is False.
 
     Returns
     -------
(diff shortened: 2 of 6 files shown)


xscen/diagnostics.py
```diff
@@ -244,8 +244,6 @@ def properties_and_measures(
         # to be able to save in zarr, convert object to string
         if "season" in ds1:
             ds1["season"] = ds1.season.astype("str")
-        if "month" in ds1:
-            ds1["month"] = ds1.month.astype("str")
```
RondeauG, line 248: @juliettelavoie Would that break the change you are implementing in your PR?

xscen/io.py
```diff
@@ -358,6 +359,7 @@ def save_to_zarr(
     encoding: dict = None,
     mode: str = "f",
     itervar: bool = False,
+    timeout_cleanup: bool = False,
```
RondeauG, line 362:
```suggestion
    timeout_cleanup: bool = True,
```

Are there cases where we don't want that? It seems to be a pretty useful QoL to prevent corrupted variables. Unless there are such situations, I'd say we either put the default as True, or just do it without adding it as a n argument to the function.

RondeauG (approved):
Looks good! Minor comments.

xscen/io.py
```diff
@@ -481,17 +487,33 @@ def coerce_attrs(attrs):
...
     else:
         logger.debug(f"Writing {list(ds.data_vars.keys())} for {filename}.")
-        return ds.to_zarr(
-            filename, compute=compute, mode="a", encoding=encoding, **zarr_kwargs
-        )
+        try:
+            return ds.to_zarr(
+                filename, compute=compute, mode="a", encoding=encoding, **zarr_kwargs
+            )
+        except TimeoutException:
```
juliettelavoie, line 509: I don't mind keeping that but it doesn't seem that useful to me, if you use a catalog..
If you follow the order  [not exist_in_catlog, xscen function, save_to_zarr, update_catalog], it doesn't matter that the file is half written, it will be overwritten.

juliettelavoie (approved):

xscen/io.py
juliettelavoie: Also, we use timeout to get out of dask being frozen. Will it even be able to do the removal or will it keep being frozen ?

aulemahal: In my case the usefulness comes from the fact that I am _appending_ to the zarr dataset. I think I am indeed not using the catalog to it's full potential. But my workflow creates the full dataset than writes it to zarr and it skips all variables already existing. This timeout cleanup ensures partially written variables are being rewritten on the second pass.

AFAIU, within the `except TimeoutException:` the fact that dask hangs is not important. The `shutil` calls do not call dask, so the files are removed even if any dask object are still "linked" to them. My usecase looks like:

 ```python
for dataset in catalog:
    with Timeout():
        with dask.Client():
            out = do_things(dataset)
            save_to_zarr(out)
```
Thus, on Timeout files are first removed, then the Client is killed and its object are garbage collected and then we go to the next iteration and we restart the counter and recreate a new client.

Comment (juliettelavoie) [reply xscen/io.py:509]: okok I understand that this way makes sens when you write a lot of indicators in one zarr.

File: xscen/diagnostics.py [hunk 4]
```diff
@@ -244,8 +244,6 @@ def properties_and_measures(
         # to be able to save in zarr, convert object to string
         if "season" in ds1:
             ds1["season"] = ds1.season.astype("str")
-        if "month" in ds1:
```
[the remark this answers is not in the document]
juliettelavoie: oh ok! mais oui je pense que ça va être correct avec le nouveau code.

PR #113, opened 2022-11-14
Regrid's weights location as kwarg

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

* In `regrid_dataset`, moves `weights_location` to the keyword argument section.

Yes, script using positional arguments for this function will break.

### Other information:
As a kwarg it can be included in the config, but I may have missed something?

### DIFF
--- a/xscen/regrid.py
+++ b/xscen/regrid.py
@@ -24,9 +24,9 @@ __all__ = ["regrid_dataset", "create_mask"]
 @parse_config
 def regrid_dataset(
     ds: xr.Dataset,
-    weights_location: Union[str, PosixPath],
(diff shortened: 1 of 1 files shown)


RondeauG:
Does that still mean that the argument is required? If not, we have to account for cases where is is None.

aulemahal:
It is still needed, but _has_ to be given as a keyword. 
In fact, my problem was that I wanted to give it in the config, while giving the datasets positionally. We could also move `weights_location` before the `,*,` and it would then be giveable both ways.

RondeauG:
I'd go with before the `*`, then.

PR #121, opened 2022-11-29
UI adjustments

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #xyz

* Adds a bit of versatility to the allowed inputs for Data/ProjectCatalogs

No.


xscen/config.py
```diff
@@ -108,6 +111,16 @@ def recursive_update(d, other):
     return d
 
 
+def args_as_str(*args: Tuple[Any, ...]) -> Tuple[str, ...]:
```
RondeauG, line 114: If the goal is to allow passing a `Path` to `DataCatalog`, wouldn't returning a tuple break it?

Zeitsperre: I'm fairly certain that the data type for a function using `*args` is `Tuple[Any, ...]`, hence why accessing `arg` from `*args` is typically done within a loop/indexing. I might be wrong, though.

RondeauG: Right, that's indeed the case.

RondeauG (approved):
I'll let you update the History and such, but otherwise it looks good! I also merged the latest 'main' with the new docs.

Issue #94, opened 2022-09-29
Improve spatial_mean


aulemahal:
### Generic Issue

While it works well, there are some caveats.

- [x] "Global" mean with xESMF : We need a way to specify we want to average over the whole grid with the xESMF method.
- [x] "cos(lat)" weighting : For the non-xESMF methods, it could be interesting to be able to weights the grid cells according to their real size with the easy `cos(lat)` approximation.
- [x] xESMF inputs : We need to re-implement the possibility to pass a `GeoDataframe` as a "shape".
- [ ] xESMF : Better error message when polygon cleaning is needed., i.e. : when `gdf.simplify(tolerance=XX, preserve_topology=True).buffer(0)` is needed. Could also be done automatically, but we'd need a way to guess the tolerance (it depends on the polygon's CRS, on the grid's resolution and on the target precision).
- [x] Segmentize polygons if needed before sending to xESMF and remove warning.
- [x] Expose xESMF's `out_chunks` arg for the regridder (xESMF 0.8)

RondeauG:
Currently, a workaround would be to use smaller polygons. For example, this uses 1 rectangle per 1° of longitude and gives the right answer:
```
eps = 1e-4
p = []
for lon in range(-180, 180):
    p.append(Polygon([(lon, -90 + eps), (lon, 90- eps), (lon + 1, 90- eps), (lon + 1, -90 + eps)]))

savg = xe.SpatialAverager(ds, p)
savg.weights.max()

savg(ds.tg_mean).mean('geom') - 273.15
```

Another solution is to circumvent `SpatialAverager` completely and use ESMF (not `xESMF`) to find the area covered by each pixel:
```
from ESMF import Field
import xesmf as xe

grid_in, _, crds = xe.frontend.ds_to_ESMFgrid(ds, need_bounds=True)
area_raw = Field(grid_in, name='area')
area_raw.get_area()

area = xr.DataArray(area_raw.data, dims=crds[::-1], coords={c: ds[c] for c in crds})

ds.tg_mean.weighted(dsa.area).mean(['lat', 'lon']) - 273.15
```

huard:
If you find bugs with xESMF, please open [issues](https://github.com/pangeo-data/xESMF/issues).

RondeauG:
Indeed, I've been talking to Pascal regarding this. We ideally want to fix the issue in xESMF, I just haven't had the time to make the Issue yet.

aulemahal:
[...]

PR #125, opened 2022-12-06
cos-lat averaging

RondeauG:
### Pull Request Checklist:
    - This PR fixes #xyz

* Added `cos-lat` method to `aggregate.spatial_mean()`
* Deprecated `mean` and added warnings to `xesmf`
* Reworked the warning messages in `_subset_file_coverage` to be more explicit
* Fixed a bug in `compute_indicators` when less than 3 timesteps were produced.

* No, but deprecates `mean` and changes the name of `interp_coord` to `interp_centroid` (still supports `interp_coord` with a deprecation warning)


RondeauG:
This is a partial fix to #94. Fixing `xesmf.SpatialAverager` is harder than expected and might require a PR in `xESMF` itself, but I didn't want to wait that long to implement cos-lat averaging.

xscen/aggregate.py
```diff
@@ -341,32 +342,75 @@ def spatial_mean(
...
+                f"{ds.attrs.get('cat:id', '')}: Latitude units is '{ds[spatial_dims['Y']].attrs['units']}', expected 'degrees_north'. "
+                f"Make sure that the computation is right."
+            )
+
+        weight = np.cos(np.deg2rad(ds[spatial_dims["Y"]]))
+        weight = weight / weight.sum()
+        ds_agg = ds.weighted(weight).mean(
+            [d for d in spatial_dims.values()], keep_attrs=True
+        )
+
```
aulemahal, line 402: I'm not sure I understand why you are doing this over the "Y" dimension and not over "latitude" directly. Like, why is 
```
weight = np.cos(np.deg2rad(ds.cf['latitude']))
weight = weight / weight.sum()
```
not correct?

AFAIU, this also support 2D grids.

You need `spatial_dims` for the effective `mean` call, but `spatial_dims["Y"]` is `rlat` in rotated pole, which is NOT wanted for the `cos-lat` method, no?

xscen/indicators.py
```diff
@@ -127,7 +127,19 @@ def compute_indicators(
...
+                        ind.injected_parameters["freq"]
+                        if "freq" in ind.injected_parameters
+                        else ind.parameters["freq"]["default"]
+                        if "freq" in ind.parameters
+                        else ind.src_freq
+                    )
+                else:
+                    freq = xr.infer_freq(out.time)
+            else:
+                freq = "fx"
```
aulemahal, line 142: Do you think ii would be better to wrap this in a private function instead of having it twice?

aulemahal (changes requested):
A few comments. I don't think the "cos-lat" think is compatible with curvilinear grids, but I'm not sure why it shouldn't be.

xscen/aggregate.py
mccrayc: Yes, we should be using 'lat' rather than 'rlat' for the weights in these cases, and I can confirm that using ds.cf['latitude'] (rather than spatial_dims['Y']) does work with 2D rotated grids. I tested this with a CRCM5 monthly statistics file with spatial_dims rlat and rlon, and the following works properly: 
```
weight = np.cos(np.deg2rad(ds.cf['latitude'])) 
da_weighted = ds['tas_moy'].weighted(weight).mean([d for d in spatial_dims.values()])
```

mccrayc (changes requested):
Two key points for the cos-lat averaging:
- No need to normalize the weights by their sum, unless I'm misunderstanding something xarray does this already.
- In agreement with what Pascal said, I believe cos-lat weighting should work fine with 2D curvilinear grids with a couple minor modifications (see comments).

aulemahal:
Je me reprends, `cos-lat` as implemented is indeed not usable with curvilinear grids. The thing is, `cos(lat)` is a weight that varies in latitude and thus assumes all cells have the same area in °² (degrees squared). Which is not true for rotated pole for example.

I found this blog : https://towardsdatascience.com/the-correct-way-to-average-the-globe-92ceecd172b7
It shows a method approximating the cell area with a lat-dependent earth radius. It however uses a cell area approximation that only works for non-curvilinear grids. You could use `cf_xarray` 0.7.6 to compute 2D bounds and then cell areas (in °).

@RondeauG your code was indeed correct. But I think curvilinear grid support is essential as one of the principal use case is with MRCC5 data.

aulemahal:
@RondeauG  I found the challenge interesting and spent a bit too much time on this, but here are 3 ways of computing cell areas for the weighted average:
https://gist.github.com/aulemahal/13c907cf2f7ccd3a84be34a7a009516c

We might not have `pygeos` already, but it's an optional dependency of `clisops`, so it might be worth adding it explicitly, it is much faster!

However, there's another issue with `cf-xarray`... Computing the 2D bounds yields strange results when the cell crosses the first lon meridian. In the MRCC5 case, the domain corner  crosses the +180 / -180 line and cell areas are super large because the corners are wrong. 

See:
image

RondeauG:
As a reference to compare to, I implemented the cos-lat method for 2D grids using `ds.cf['latitude']` instead of `ds.cf['X']`. Here's an example for RegCM4 / NAM-22:

image

RondeauG:
I'll look into the links provided, though I feel like this might be another method entirely? As in, we'd have `cos-lat` or `SpatialAverager` or `pygeos/ESMF/whatever`?

RondeauG:
Also, with the changes, `get_spatial_dims(ds)` is pretty useless. I could restore it within `method=="mean"` as it was before, but I'm tempted to just get rid of `mean` entirely since it does the wrong thing to begin with.

aulemahal:
> this might be another method entirely?
My "shapely" and "pygeos" functions are both based on the cos-lat approximation, so I would argue it's just a curvilinear-compatible version of `cos-lat`.

The ESMF version is indeed another technique because it computes the cell area using other approximations. 

Note for shapely:  the newest version (2.0) now merges the functionality of `pygeos` so there might be a way to code this idea without adding a new dependency.

aulemahal:
I updated my gist with a function that works with shapely 2.0. It is indeed the exact same as with `pygeos` and it's as fast.

RondeauG:
@aulemahal @mccrayc 
This might be a lack of comprehension from my part on rotated grids, but let's say that I have this region:

image

The first part of Pascal's code (using the `pygeos` version for example) produces this:
`xr.DataArray(pygeos.area(pygeos.polygons(pygeos.linearrings(ds.lon_bounds, ds.lat_bounds))), dims=ds.lon.dims, coords=ds.lon.coords).plot()`

image

`np.cos(np.deg2rad(ds.lat))` creates more or less the opposite:

image

Meaning that when you multiply both, you get:

image

...which makes no intuitive sense to me? I would have expected weights to be curved, not linear like that?

aulemahal:
No I think this is ok! You are in a Oblique  Mercator projection, which projects the globe unto a cylinder. The projection being centered at y = 0, in the center of your plot, it makes sense to me that the cell area decreases with `abs(y)`.

https://en.wikipedia.org/wiki/Oblique_Mercator_projection

Also, I found a way to compute areas (in meters) with `pyproj` from lat and lon bounds. This was added to the gist, and gives the following:
image

xscen/aggregate.py
```diff
@@ -334,36 +335,70 @@ def spatial_mean(
...
+            logger.warning(
+                f"{ds.attrs.get('cat:id', '')}: Latitude does not appear to have units. Make sure that the computation is right."
+            )
+        elif ds.cf["latitude"].attrs["units"] != "degrees_north":
+            logger.warning(
+                f"{ds.attrs.get('cat:id', '')}: Latitude units is '{ds.cf['latitude'].attrs['units']}', expected 'degrees_north'. "
+                f"Make sure that the computation is right."
+            )
+
+        if (ds.cf["longitude"].min() < -160) & (ds.cf["longitude"].max() > 160):
```
aulemahal, line 373:
```suggestion
        if (
            ((ds.cf["longitude"].min() < -160) & (ds.cf["longitude"].max() > 160))
            or ((ds.cf["longitude"].min() < 20) & (ds.cf["longitude"].max() > 340))
        ):
```
To catch the other edge case.

xscen/aggregate.py
```diff
@@ -334,36 +335,70 @@ def spatial_mean(
...
+        ) * np.cos(np.deg2rad(ds.cf["latitude"]))
+
+        ds_agg = ds.weighted(weights).mean(
+            [d for d in ds.cf.axes["X"] + ds.cf.axes["Y"]], keep_attrs=True
+        )
+
+        # Prepare the History field
+        new_history = (
+            f"[{datetime.datetime.now().strftime('%Y-%m-%d %H:%M:%S')}] "
+            f"weighted mean(dim={[d for d in ds.cf.axes['X'] + ds.cf.axes['Y']]}) using a 'cos-lat' approximation"
```
aulemahal, line 395: I'm not sure how to phrase it, but I feel this should mention that we are also weighting by the cell area in deg².

aulemahal (approved):
This looks good!

Semi-related:
As we currently have no bulletproof method, the choice makes sense. However, it seems to me that there's no reason to have such an array of methods for such a simple task. I guess that when we have a simple, light area-weighted mean method, we could remove everything else? Is there a use for a collection of imperfect algos ?

This would mean :
- fix bounds generation when crossing meridians in `cf_xarray`
- generate `areacella` in m² from those bounds (using `pyproj` and `shapely`)

RondeauG:
> Semi-related: As we currently have no bulletproof method, the choice makes sense. However, it seems to me that there's no reason to have such an array of methods for such a simple task. I guess that when we have a simple, light area-weighted mean method, we could remove everything else? Is there a use for a collection of imperfect algos ?

That's why I deprecate `mean`, since `cos-lat` is a superior version of the same thing. I could see an argument to drop `interp_centroid` in favor of always using `xesmf` too. 

However, I think that `cos-lat` and `xesmf` (SpatialAverager) fundamentaly do different things. `cos-lat` keeps the grid cells as is, while `SpatialAverager` will break pixels into smaller parts to closely fit to a given polygon.

xscen/aggregate.py
RondeauG: `using a 'cos-lat' approximation of areacella (in deg2)` ?

aulemahal: Agreed!

aulemahal:
@mccrayc When you have time, a second review is needed since you "requested changes".

mccrayc (approved):
Looks good to me!

xscen/aggregate.py
```diff
@@ -413,6 +455,10 @@ def spatial_mean(
 
     # Uses xesmf.SpatialAverager
     elif method == "xesmf":
+        logger.warning(
+            "A bug has been found with xesmf.SpatialAverager that impacts big or regions. "
```
RondeauG, line 459:
```suggestion
            "A bug has been found with xesmf.SpatialAverager that appears to impact big regions. "
```

xscen/indicators.py
```diff
@@ -144,6 +165,11 @@ def compute_indicators(
                 concats.extend(tmp)
             out = xr.concat(concats, dim="time")
 
+        # Make sure that attributes have been kept for the dimensions and coordinates. Fixes a bug in flox.
```
RondeauG, line 168:
```suggestion
        # Make sure that attributes have been kept for the dimensions and coordinates. Fixes a bug in xarray.
```

xscen/aggregate.py
```diff
@@ -334,36 +335,72 @@ def spatial_mean(
...
+        warnings.warn(
+            "xs.spatial_mean with method=='mean' is deprecated and will be abandoned in a future release. Use method=='cos-lat' instead for a more robust but similar method.",
+            category=DeprecationWarning,
+            stacklevel=2,
+        )
+    elif method == "interp_coord":
+        warnings.warn(
+            "xs.spatial_mean with method=='interp_coord' is deprecated. Use method=='interp_centroid' instead.",
+            category=DeprecationWarning,
+            stacklevel=2,
```
aulemahal, line 348:
```suggestion
            category=FutureWarning,
        )
    elif method == "interp_coord":
        warnings.warn(
            "xs.spatial_mean with method=='interp_coord' is deprecated. Use method=='interp_centroid' instead.",
            category=FutureWarning,
```
From python's doc:

**DeprecationWarning** : Base category for warnings about deprecated features when those warnings are intended for other Python developers (ignored by default, unless triggered by code in __main__).
**FutureWarning** : Base category for warnings about deprecated features when those warnings are intended for end users of applications that are written in Python.

Also, `stacklevel=2` lands in the `parse_config` wrapper. Without it (=1), the warning is shown was stemming from inside the function. And with 3, it stems from call of `spatial_mean` (like the workflow or the notebook cell itself).

PR #127, opened 2022-12-09
Faster search

aulemahal:
### Pull Request Checklist:
    - This PR fixes #
-  [x] (If applicable) Documentation has been added / updated (for bug fixes / features)

### The issue
`search_data_catalogs` and `extract_dataset` are very slow when the catalogs are very big. The base case for this PR was raised by @coxipi and replicated by me : `search_data_catalog ` over the MRCC5, with a selection that returned 0 datasets, took 12 min. The same process, but coded through `DataCatalog.search()` took 2 min.

Faster `search_data_catalogs` and `extract_dataset` through:

* Faster `DataCatalog.unique`: I copied the `cat.unique()` implementation from `intake-esm` but instead of computing ALL unique values and then extracting those from the column(s) I want, I _only_ compute for those columns. 
* Remove logging call that used `DataCatalog.nunique()` : for the same reason as above, instead of logging :`catalog.nunique()['id']`, I switched to `len(catalog.unique()['id'])`. That made an improvement of more than 5 min. LOL.
* Faster date parsing : yet another implementation of our custom `parse_dates` to optimize the case when we have dates out of the `datetime64[ns]` bounds.  I went from 50 s to 3 s for `simulation.json`.
* Rewrite of the ``ensure_correct_time`` logic. This one is funny. Did you know that `any(da > 1)` is 10x slower than `(da > 1).any()` . I didn't. I also added a fast-track for cases where `infer_freq` works (most of them duh). 

I ran the `getting_started` notebook and got no error. CQFD,

Seriously, I don't think so. The error raised when we have invalid date strings in catalogs may have changed, but it is still explicit.

### Other information:
The `unique()` improvement could be moved to `intake_esm` but I don't have the energy.


xscen/extract.py
```diff
@@ -261,6 +261,29 @@ def extract_dataset(
...
         ):
+            if "time" in ds_ts and ensure_correct_time:
+                # Expected freq (xrfreq is the wanted freq)
+                expfreq = catalog[key].df.xrfreq.iloc[0]
+                # Check if we got the expected freq (skip for too short timeseries)
+                inffreq = xr.infer_freq(ds_ts.time) if ds_ts.time.size > 2 else None
+                if inffreq == expfreq:
+                    # Even the freq is correct, we ensure the correct "anchor" for daily and finer
+                    if expfreq in "DHTMUL":
+                        ds_ts["time"] = ds_ts.time.dt.floor(expfreq)
```
RondeauG, line 272: I don't think this fixes the issue of variables that are not quite at the same hour?

xscen/catalog.py
```diff
@@ -118,18 +119,28 @@ def _parse_list_of_strings(elem):
...
+#         return pd.PeriodIndex(elem, freq="H")
 def _parse_dates(elem):
-    try:
-        if isinstance(elem, str):
-            return pd.Timestamp(elem).to_period("H")
-        return pd.DatetimeIndex(elem).to_period("H")
-    except pd.errors.OutOfBoundsDatetime:
-        warnings.warn(
-            "Somes dates are out of the datetime64[ns] range, switching to slower direct period parsing."
-        )
```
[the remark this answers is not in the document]
aulemahal: I'm not sure I see the advantage because now there is no switch, it always go through the same code. Also, `time.isnull()` also picks up `fx` entries, so the `PeriodIndex` call is almost always non-empty, but that doesn't mean that dates are out of the range.

I would prefer document it elsewhere in the doc than raise a warning each time we read a catalog.

xscen/extract.py
aulemahal: It doesn't fix the issue where different variables of the same dataset (after `to_dataset_dict`) have different hours, indeed. But the previous code didn't either. That case will trigger the first `ValueError` below.

xscen/catalog.py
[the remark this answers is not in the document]
RondeauG: Alright, good with me!

xscen/extract.py
RondeauG: Alright. I can approve this PR as is and we'll look into that at a later date?

aulemahal: Yep! It wasn't my plan here to fix this, only to accelerate it.

aulemahal:
@RondeauG do you have an idea about the failures in my tests? It seems to be the "ensemble reduction" notebook that doesn't work as expected...

aulemahal:
Woups my bad forget it. It's from a change I made in `DataCatalog.unique()`

RondeauG:
> Woups my bad forget it. It's from a change I made in `DataCatalog.unique()`

Yeah, I think `xrfreqs=ds_dict.unique("xrfreq"),` might be crashing because you changed the type of the output?

RondeauG (approved):

PR #138, opened 2023-01-03
add restrict_wl

juliettelavoie:
### Pull Request Checklist:
    - This PR fixes #xyz

*  new argument to restrict the datasets to only ones that exist in the warming level csv. 
*  `subset_warming_level` fails if the dataset doesn't exist. Hence, filtering beforehand is necessary.
* It is possible to pass a boolean or a dictionnary with arguments similar to `subset_warming_level` (ignore_member and tas_csv).

 no

### DIFF
--- a/xscen/extract.py
+++ b/xscen/extract.py
@@ -551,7 +551,7 @@ def search_data_catalogs(
         Used to restrict the results to a given number of members for a given simulation.
         Currently only supports {"ordered": int} format.
     restrict_warming_level : bool,dict
-        Used to restrict the results to only datasets that exist in the csv used to compute warming levels.
+        Used to restrict the results to only datasets that exist in the csv used to compute warming levels in `subset_warming_level`.
         If True, this will only keep the datasets that have a mip_era, source, experiment
         and member combination that exist in the csv. This does not guarantees that a given warming level will be reached, only that the datasets have corresponding columns in the csv.
         More option can be added by passing a dictionary instead of a boolean.
--- a/xscen/extract.py
+++ b/xscen/extract.py
@@ -514,6 +514,7 @@ def search_data_catalogs(
     conversion_yaml: str = None,
     restrict_resolution: str = None,
     restrict_members: dict = None,
+    restrict_warming_level: dict = None,
 ) -> dict:
     """Search through DataCatalogs.
 
@@ -549,6 +550,13 @@ def search_data_catalogs(
     restrict_members : dict
         Used to restrict the results to a given number of members for a given simulation.
         Currently only supports {"ordered": int} format.
+    restrict_warming_level : bool,dict
+        Used to restrict the results to only datasets that exist in the csv used to compute warming levels.
+        If True, this will only keep the datasets that have a mip_era, source, experiment
+        and member combination that exist in the csv. This does not guarantees that a given warming level will be reached, only that the datasets have corresponding columns in the csv.
+        More option can be added by passing a dictionary instead of a boolean.
+        If {'ignore_member':True}, it will keep dataset that have the right mip_era, source and experiment, regardless of member.
+        If {tas_csv: Path_to_csv}, it will use this csv instead of the default one provided by xscen.
 
     Notes
     -----
@@ -777,6 +785,15 @@ def search_data_catalogs(
 
     if restrict_members is not None and len(catalogs) > 0:
         catalogs = _restrict_multimembers(catalogs, id_columns, restrict_members)
+
+    if restrict_warming_level and len(catalogs) > 0:
+        if isinstance(restrict_warming_level, bool):
+            restrict_warming_level = {}
+        if "ignore_member" not in restrict_warming_level:
+            restrict_warming_level["ignore_member"] = False
+        if "tas_csv" not in restrict_warming_level:
+            restrict_warming_level["tas_csv"] = None
+        catalogs = _restrict_wl(catalogs, restrict_warming_level)
 
     return catalogs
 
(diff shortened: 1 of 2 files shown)


xscen/extract.py
```diff
@@ -560,6 +561,13 @@ def search_data_catalogs(
     restrict_members : dict
         Used to restrict the results to a given number of members for a given simulation.
         Currently only supports {"ordered": int} format.
+    restrict_warming_level : bool,dict
+        Used to restrict the results to only datasets that exist in the csv used to compute warming levels in `subset_warming_level`.
```
RondeauG, line 565:
```suggestion
        Used to restrict the results only to datasets that exist in the csv used to compute warming levels in `subset_warming_level`.
```

xscen/extract.py
```diff
@@ -560,6 +561,13 @@ def search_data_catalogs(
...
+        If True, this will only keep the datasets that have a mip_era, source, experiment
+        and member combination that exist in the csv. This does not guarantees that a given warming level will be reached, only that the datasets have corresponding columns in the csv.
+        More option can be added by passing a dictionary instead of a boolean.
+        If {'ignore_member':True}, it will keep dataset that have the right mip_era, source and experiment, regardless of member.
```
RondeauG, line 569:
```suggestion
        If {'ignore_member':True}, it will disregard the member when trying to match the dataset to a column.
```

xscen/extract.py
```diff
@@ -560,6 +561,13 @@ def search_data_catalogs(
...
+        If {tas_csv: Path_to_csv}, it will use this csv instead of the default one provided by xscen.
```
RondeauG, line 570:
```suggestion
        If {tas_csv: Path_to_csv}, it will use an alternative csv instead of the default one provided by xscen.
```

xscen/extract.py
```diff
@@ -798,6 +806,15 @@ def search_data_catalogs(
     if restrict_members is not None and len(catalogs) > 0:
         catalogs = _restrict_multimembers(catalogs, id_columns, restrict_members)
```
RondeauG, line 808: Is there a reason to do this on `catalogs` instead of doing it way earlier on `catalog` (after `if exclusions`, for example)? The way this is currently implemented, you potentially lose a lot of time searching through simulations that you might end up deleting anyway.

At the very least, this should come before `restrict_resolution` and `restrict_members`, since these should be the last checkups (aka you've done all your work, and only then do you eliminate duplicates if some still exist).

xscen/extract.py
```diff
@@ -798,6 +806,15 @@ def search_data_catalogs(
...
 
+    if restrict_warming_level and len(catalogs) > 0:
+        if isinstance(restrict_warming_level, bool):
+            restrict_warming_level = {}
+        if "ignore_member" not in restrict_warming_level:
+            restrict_warming_level["ignore_member"] = False
+        if "tas_csv" not in restrict_warming_level:
+            restrict_warming_level["tas_csv"] = None
```
RondeauG, line 815:
```suggestion
        restrict_warming_level = restrict_warming_level.setdefault("ignore_member", False)
        restrict_warming_level = restrict_warming_level.setdefault("tas_csv", None)
```

xscen/extract.py
```diff
@@ -1223,6 +1240,36 @@ def _restrict_multimembers(catalogs: dict, id_columns: list, restrictions: dict)
     return catalogs
 
 
+def _restrict_wl(catalogs: dict, restrictions: dict):
+    """Update the results from search_data_catalogs by removing simulations that are not available in the warming level csv."""
+    df = pd.concat([catalogs[s].df for s in catalogs.keys()])
```
RondeauG, line 1245: Following my other comment on `catalog` instead of `catalogs`, this line won't be needed.

xscen/extract.py
```diff
@@ -1223,6 +1240,36 @@ def _restrict_multimembers(catalogs: dict, id_columns: list, restrictions: dict)
...
+    else:
+        df["csv_name"] = df["mip_era"].str.cat(
+            [df["source"], df["experiment"], df["member"]], sep="_"
+        )
+        csv_source = list(annual_tas.columns[1:])
+
+    to_remove = pd.unique(df[~df["csv_name"].isin(csv_source)]["id"])
+
+    for k in to_remove:
+        logger.info(f"Removing {k} from the results.")
```
RondeauG, line 1267: If you follow my suggestion (`catalog` vs `catalogs`), I would remove this `logger` since it will potentially spam thousands of lines.

Also using `catalog`, I'd make `to_remove` a Series (or single row dataframe?) of booleans and subset based on that, which would be much faster than a `for` loop. Something like https://stackoverflow.com/questions/29281815/pandas-select-dataframe-columns-using-boolean

RondeauG:
I also don't see changes to the documentation? We should add a mention of this in the Catalogs notebook, where we list possible options: https://xscen.readthedocs.io/en/latest/notebooks/1_catalog.html#Advanced-search:-xs.search_data_catalogs

xscen/extract.py
juliettelavoie: There is no reason, I just figured it was the best way to exclude stuff as this is what had been done for other restrictions. I can put it after `if exclusion`.

xscen/extract.py
juliettelavoie: Careful here. We can do `restrict_warming_level.setdefault("ignore_member", False)` only. If we do `restrict_warming_level =` it will change the dictionnary into a boolean.

xscen/extract.py
juliettelavoie: Ok for subsetting  without the for loop. But, I think it would still be important to log the unique ids that are being removed.

RondeauG (approved):
Looks good!

PR #143, opened 2023-01-24
fix for functions with decorators in Pycharm 2022

RondeauG:
### Pull Request Checklist:
    - This PR fixes #xyz

* This fixes a bug in Pycharm 2022 CE, where hints for decorated functions are reduced to `(*args, **kwargs)` instead of actual relevant information. This bug is quite annoying, since this is 90% of `xscen`.
* Issue on JetBrains: https://youtrack.jetbrains.com/issue/PY-48338
* The fix was taken from here (without the need to install that package): https://pypi.org/project/decohints/ 

Probably not? @aulemahal is the one who understands `config.py` 😅


aulemahal (approved):
This is only an annotation, it should be harmlesss.

PR #147, opened 2023-01-31
Custom aggregation

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

New `to_dataset` method on `DataCatalog`. 

Same as `to_dask`, but exposes options to change the aggregation control :
* `concat_on` to list columns over which the datasets are concatenated.
* `ensemble_on` to list columns over which a `realization` dimension is created.

The goal of this function is to reduce code complexity is some common cases where one wants a dataset with all members and experiments (for examples).

I also added a "good to know" page to the doc. A place where to list all sorts of misc information that users of xscen should be aware of. The first section is about how to open data.

### Example:
```python3
cat = xs.DataCatalog("/you/know/where/ESPO-extra.json")

ds = cat.search(
    bias_adjust_project='ScenGen', xrfreq='QS-DEC'
).to_dataset(concat_on=['experiment'], ensemble_on=['institution', 'source'])
```
The output:
```
<xarray.Dataset>
Dimensions:                   (experiment: 2, realization: 11, time: 605, lat: 320, lon: 416)
Coordinates:
  * lat                       (lat) float32 66.62 66.54 66.46 ... 40.12 40.04
  * lon                       (lon) float32 -89.05 -88.96 ... -54.55 -54.46
  * time                      (time) datetime64[ns] 1949-12-01 ... 2100-12-01
  * experiment                (experiment) object 'rcp45' 'rcp85'
  * realization               (realization) object 'CCCma_CanESM2' ... 'NOAA-...
...
```

The same code with pure xscen:
```python3
dss = []
for exp in ['rcp45', 'rcp85']:
    cats = xs.search_data_catalogs(
        "/you/know/where/ESPO-extra.json",
        variables_and_freqs={ind: 'QS-DEC' for ind in indicators},
        other_search_criteria={'domain': 'QC'},
    )
    dsd = {}
    for dsid, cat in cats.items():
        dsd[dsid] = xs.extract_dataset(cat)['QS-DEC']
    dss.append(xclim.ensembles.create_ensembles(dsd, calendar='standard', resample_freq='QS-DEC')
ds = xr.concat(dss, xr.DataArray(['rcp45', 'rcp85'], dims=('experiment',), name='experiment'))
```

The plus value of this PR seems evident to me here.

### Where ?
As it was developed for an ensemble case, I implemented it as a single-dataset output. But that seems a bit limited. In a xscen-world, this could be implemented at the `search_data_catalogs` level. 

However, it felt to me that the `search_data_catalogs`/`extract_dataset` combo is best used for raw data. Once at the ensemble step, simples `DataCatalog.search` are often enough. Thus, the need to have this on `DataCatalog`.

Also, this could be moved upstream to `intake-esm`, but the `ensemble_on` and `calendar` args seem a bit to xclim-specific...


xscen/catalog.py
```diff
@@ -347,6 +347,77 @@ def exists_in_cat(self, **columns):
             logger.info(f"An entry exists for: {columns}")
         return exists
 
+    def to_dataset(
```
juliettelavoie, line 350: Could you write some examples because this is not performing the way I am expecting from the docstring?
 
 2 examples of it not doing what I expected:
 
 1. `cat_with_4_experiments.to_dataset(concat_on=['experiment'])` I am expecting this to work and have a coordinate 'experiment'.
 
 What I get:  `ValueError: Expected exactly one dataset. Received 4 datasets.`
 
 2. `cat_with_4_experiments_and_3_members.to_dataset(ensemble_on=[ 'experiment'])`. I am expecting this to fail because we have too many members.
 
 What I get: I works and returns a dataset with coordinate `realization` with the 4 experiments. Unclear which member was used.

juliettelavoie (commented):
I think the good to know page is really neat!
I am having issues using `to_dataset`. Unclear to me if I don't understand what it is suppose to do or it is not working...

docs/goodtoknow.rst
```diff
@@ -0,0 +1,45 @@
...
+
+Which function to use when resampling data
+------------------------------------------
+
+:extract_dataset: :py:func:`~xscen.extract.extract_dataset` has resampling capabilities to provide daily data from finer sources.
+
+:xclim indicators: Through :py:func:`~xscen.indicators.compute_indicators`, xscen workflows can easily use `xclim indicators <https://xclim.readthedocs.io/en/stable/indicators.html>`_
+	to go from daily data to coarser (monthly, seasonal, annual).
+
+What is missing in `xscen` and `xclim` is thus a good way to start the resampling from data coarser than daily, where the base period is non-uniform (ex: months have different lengths).
```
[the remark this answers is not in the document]
aulemahal: I simply wanted to highlight that  xscen/xclim doesn't provide a clean way to resample data coarser than daily. For example, going from monthly to yearly. The simple xarray mechanism doesn't understand that months have different lengths and should be weighted differently in the resampling mean.

xscen/catalog.py
aulemahal: Hum woups, I think this is because I don't handle the `id` correctly below. Will change and make more tests.

aulemahal: Last commit should have fixed the issue.

```python3
import xscen as xs
cat = xs.DataCatalog('catalogues/simulation.json')

scat = cat.search(mip_era='CMIP6', variable='tas', source='FGOALS.*', activity='ScenarioMIP')
scat.to_dataset(concat_on='experiment', ensemble_on='source')
```
works

juliettelavoie: Ok example 1 works as expected now!
But, example 2 still doesn't fail...
With the simulationc catalogue:
`cat.search(source='MIROC6', xrfreq='D', activity='ScenarioMIP').to_dataset(ensemble_on=['experiment'])`

The subcat has 12 datasets (3 members x 4 experiments), but the output only has 4 datasets inside it (the experiments). I would have expected this to fail...

docs/goodtoknow.rst
```diff
@@ -0,0 +1,45 @@
...
+	``extract_dataset`` on each item. This then returns a dictionary with a single dataset for each ``xrfreq``. You thus end up with one dataset per frequency and ``id``.
+
+:to_dataset_dict: Using :py:meth:`~xscen.catalog.DataCatalog.to_dataset_dict`.
+    When all the data you need is in a single catalog (for example, your :py:meth:`~xscen.catalog.ProjectCatalog`) and you don't need any of the features listed above.
+    As explained in :doc:`columns`, it creates a dictionary with a single Dataset for each combination of ``id``, ``domain``, ``processing_level`` and ``xrfreq``.
+
+:to_dataset: Using :py:meth:`~xscen.catalog.DataCatalog.to_dataset`.
+    Similar to ``to_dataset_dict``, but only returns a single dataset. If the catalog has more than one, the call will fail. However, it exposes options to add aggregations.
+    This is useful when constructing an ensemble dataset that would otherwise result in distinct entries in the output of ``to_dataset_dict``. It can usually be used in
+    replacement of a combination of ``to_dataset_dict`` and :py:func:`~xclim.ensembles.create_ensemble`.
```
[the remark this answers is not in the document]
aulemahal: All instances changed in the notebooks.

xscen/catalog.py
aulemahal: Ok, so my ID editing logic was not good. I instead _added_ a `realization` column to concat on and edited the ID by removing all concatenated files. It should now work!

docs/goodtoknow.rst
RondeauG, line 45:
```suggestion
What is currently not covered by either `xscen` or `xclim` is a method to resample from data coarser than daily, where the base period is non-uniform (ex: resampling from monthly to annual data, taking into account the number of days per month).
```

xscen/catalog.py
```diff
@@ -347,6 +347,113 @@ def exists_in_cat(self, **columns):
...
+                        type=intake_esm.cat.AggregationType.join_new, attribute_name=col
+                    )
+                    for col in concat_on
+                ]
+            )
+
+        if ensemble_on:
+            if preprocess is not None:
+                warnings.warn(
+                    "Using `ensemble_on` will override the iven `preprocess` function."
```
RondeauG, line 426:
```suggestion
                    "Using `ensemble_on` will override the given `preprocess` function."
```

xscen/catalog.py
```diff
@@ -347,6 +347,113 @@ def exists_in_cat(self, **columns):
...
+        Open the catalog's entries into a single dataset.
+
+        Same as :py:meth:`~intake_esm.core.esm_datastore.to_dask`, but with additionnal control over the aggregations.
+        Ensemble preprocessing logic is taken from :py:func:`xclim.ensembles.create_ensemble`.
+        When `ensemble_on` is given, the function ensures all entries have the correct time coordinate according to `xrfreq`.
+
+        Parameters
+        ----------
+        concat_on : list of strings or str, optional
+          A list of catalog columns to concat the datasets over. Each will become a new dimension with the column values as coordinates.
```
RondeauG, line 367: I thought I understood what `to_dataset` did until I read this line...

`additionnal control over the aggregations` sounded to me like we could change the aggregation controls, as is implied in the text and in the code, but as far as I know, this does not add dimensions to the dataset? It's more of a way to say `"These combination of columns define which files can be opened at the same time"`, and then you have to use `xarray_combine_by_coords_kwargs` if dimensions do not match.

juliettelavoie: It does add dimension. For me (and maybe for you), the use of "aggregation" was confusing. Elsewhere in xscen, aggregation means average and reducing the time or space dimension. Here, it means adding new dimension to the dataset, similar to create_ensemble (aggregating many datasets in one).

RondeauG: My bad, I confused `aggregation_control.aggregations` with `aggregation_control.groupby_attrs`. That being said, I think the name and description is kind of confusing. I'll think about a suggestion.

xscen/catalog.py
```diff
@@ -347,6 +347,115 @@ def exists_in_cat(self, **columns):
...
+        Same as :py:meth:`~intake_esm.core.esm_datastore.to_dask`, but with additionnal control over the aggregations.
+        Ensemble preprocessing logic is taken from :py:func:`xclim.ensembles.create_ensemble`.
+        When `ensemble_on` is given, the function ensures all entries have the correct time coordinate according to `xrfreq`.
+        If either argument is given, the "id" is reconstructed by removing mentions of aggregated columns.
+        This will override any "custom" id, ones not unstackable with :py:func:`~xscen.catalog.unstack_id`.
+
+        Parameters
+        ----------
+        concat_on : list of strings or str, optional
+          A list of catalog columns to concat the datasets over. Each will become a new dimension with the column values as coordinates.
```
RondeauG, line 369:
```suggestion
          A list of catalog columns over which to concat the datasets (in addition to 'time'). Each will become a new dimension with the column values as coordinates.
```

xscen/catalog.py
```diff
@@ -347,6 +347,115 @@ def exists_in_cat(self, **columns):
...
+        self,
+        concat_on: Optional[Union[List[str], str]] = None,
+        ensemble_on: Optional[Union[List[str], str]] = None,
+        calendar: Optional[str] = "standard",
+        **kwargs,
+    ) -> xr.Dataset:
+        """
+        Open the catalog's entries into a single dataset.
+
+        Same as :py:meth:`~intake_esm.core.esm_datastore.to_dask`, but with additionnal control over the aggregations.
```
RondeauG, line 360:
```suggestion
        Same as :py:meth:`~intake_esm.core.esm_datastore.to_dask`, but with additional control over the aggregations. The dataset definition logic is left untouched by this method (by default: ["id", "domain", "processing_level", "xrfreq"]). What this method does is provide a way to concatenate datasets from the catalog itself, using `intake_esm`'s functionalities. `concat_on` will keep the column names and add them as new dimensions, while `ensemble_on` will override the column names and concatenate the datasets along a new `realization` dimension. Both can be used at the same time.
        
```

xscen/catalog.py
```diff
@@ -347,6 +347,115 @@ def exists_in_cat(self, **columns):
...
+        When `ensemble_on` is given, the function ensures all entries have the correct time coordinate according to `xrfreq`.
+        If either argument is given, the "id" is reconstructed by removing mentions of aggregated columns.
+        This will override any "custom" id, ones not unstackable with :py:func:`~xscen.catalog.unstack_id`.
+
+        Parameters
+        ----------
+        concat_on : list of strings or str, optional
+          A list of catalog columns to concat the datasets over. Each will become a new dimension with the column values as coordinates.
+          Xarray concatenation rules apply and can be acted upon through `xarray_combine_by_coords_kwargs`.
+        ensemble_on : list of strings or str, optional
```
RondeauG, line 371:
```suggestion
        create_ensemble_on : list of strings or str, optional
```
Maybe? To match the xclim name (since this is essentially what this does).

RondeauG: Also, a more natural logic for this argument (to me at least) would be something like : `create_ensemble_on: ['member']` (where we then remove the member from `id` and create the ensemble), instead of having to list all the columns except the one that actually matters.

xscen/catalog.py
```diff
@@ -347,6 +347,115 @@ def exists_in_cat(self, **columns):
...
+
+        Same as :py:meth:`~intake_esm.core.esm_datastore.to_dask`, but with additionnal control over the aggregations.
+        Ensemble preprocessing logic is taken from :py:func:`xclim.ensembles.create_ensemble`.
+        When `ensemble_on` is given, the function ensures all entries have the correct time coordinate according to `xrfreq`.
+        If either argument is given, the "id" is reconstructed by removing mentions of aggregated columns.
+        This will override any "custom" id, ones not unstackable with :py:func:`~xscen.catalog.unstack_id`.
+
+        Parameters
+        ----------
+        concat_on : list of strings or str, optional
```
RondeauG, line 368:
```suggestion
        extra_concat_on : list of strings or str, optional
```
My issue here is that people might try to say `concat_on: 'time'` or something like that. It should be more obvious in the name that this is in addition to 'time'. I'm not a fan of `extra`, just using it as an example.

xscen/catalog.py
juliettelavoie: Personnally, I like the "put the columns that you want" logic better.

docs/goodtoknow.rst
```diff
@@ -0,0 +1,45 @@
...
+	* variable conversion and resampling
+	* spatial and temporal subset
+	* matching historical and future runs for simulations
+
+	``search_data_catalogs``  returns a dictionary with a specific catalog for each of the unique ``id`` found in the search. One should then iterate over this dictionary and call
+	``extract_dataset`` on each item. This then returns a dictionary with a single dataset for each ``xrfreq``. You thus end up with one dataset per frequency and ``id``.
+
+:to_dataset_dict: Using :py:meth:`~xscen.catalog.DataCatalog.to_dataset_dict`.
+    When all the data you need is in a single catalog (for example, your :py:meth:`~xscen.catalog.ProjectCatalog`) and you don't need any of the features listed above.
+    As explained in :doc:`columns`, it creates a dictionary with a single Dataset for each combination of ``id``, ``domain``, ``processing_level`` and ``xrfreq``.
```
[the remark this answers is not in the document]
aulemahal: I think it's ok to be there. But at one point, this should be documented in some way...

xscen/catalog.py
aulemahal: Hum. I understand, but I felt the concatenation over time was so self-evident that this misunderstanding (passing `concat_on=['time']`) wouldn't really be an issue. I think your additions to the docstring should be enough?

xscen/catalog.py
aulemahal: Not sure I understand, @RondeauG , aren't you describing what this function does?

The difference with `concat_on` is that `ensemble_on` will merge the given column values into a new id-like column. Then it adds an aggregation over that new column and removes the given columnd from the dataset id, so as to remove them from the `groupby_attrs` logic. 

I have no hard feelings against `create_ensemble_on`.

aulemahal:
@RondeauG I re-edited the docstring because it felt a bit redundant.

xscen/catalog.py
```diff
@@ -347,6 +347,117 @@ def exists_in_cat(self, **columns):
...
+        Ensemble preprocessing logic is taken from :py:func:`xclim.ensembles.create_ensemble`.
+        When `create_ensemble_on` is given, the function ensures all entries have the correct time coordinate according to `xrfreq`.
+
+        Parameters
+        ----------
+        concat_on : list of strings or str, optional
+          A list of catalog columns over which to concat the datasets (in addition to 'time'). Each will become a new dimension with the column values as coordinates.
+          Xarray concatenation rules apply and can be acted upon through `xarray_combine_by_coords_kwargs`.
+        create_ensemble_on : list of strings or str, optional
+          The given column values will be merged into a new id-like "realization" column, which will concatenated over.
```
RondeauG, line 374:
```suggestion
          The given column values will be merged into a new id-like "realization" column, which will be concatenated over.
```

RondeauG (approved):
I think it's good to go!

PR #167, opened 2023-03-14
Add measures_improvement_2d

juliettelavoie:
### Pull Request Checklist:
    - This PR fixes #xyz

*  Allows to create a 2d dataset showing the number of improved grid cells. It can either concat improvements already calculated or calculate the improvements for many datasets by providing their measures.

no

### Other information:
My goal is to easily produce this 
image
instead of a bunch of these
image

### DIFF
--- a/xscen/diagnostics.py
+++ b/xscen/diagnostics.py
@@ -13,7 +13,7 @@ from .catalog import DataCatalog
 from .config import parse_config
 from .indicators import load_xclim_module
 from .io import save_to_zarr
-from .utils import change_units, unstack_fill_nan
+from .utils import change_units, clean_up, unstack_fill_nan
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


xscen/diagnostics.py
```diff
@@ -370,3 +370,41 @@ def measures_improvement(
     ds_better.attrs.pop("cat:variable", None)
 
     return ds_better
+
+
+def measures_improvement_2d(dict_input: dict, to_level: str = "diag-improved-2d"):
+    """
+    Create a 2D dataset with dimension `datasets` showing the fraction of improved grid cell.
```
RondeauG, line 377: Should we use `realization` instead of `datasets`, since this is the name we usually give to that dimension?

RondeauG (approved):

xscen/diagnostics.py
juliettelavoie: ahh that's a good idea. I was copying the format of `measures_heatmap`. So, we should probably change it there too, but that would be a breaking change...

RondeauG: I don't know if many people use `measures_heatmap`. We could maybe roll out a FutureWarning and eventually change that function too?

juliettelavoie: How does that work ? I just say: "In a future version, the `datasets` dimension will be renamed `realization`."  How long do we wait before doing it ?

RondeauG: Something similar to what I've done here: https://github.com/Ouranosinc/xscen/blob/main/xscen/aggregate.py#L354.

As for the timing, I think the best practice would be to make the warning now and keep it until `0.6.x`, then make the change in `0.7.0` ? @Zeitsperre

Zeitsperre: Yes sir.
 - Emit a `FutureWarning` that the function/call signature will be removed in a future version.
 - Point users to the release notes in the deprecation notice in the warning message.
 - Effect the change in two minor versions (~2 months).

juliettelavoie: but actually it's not the signature that is changing. It's the name of the dimension of the output. So maybe it's less a big deal? maybe "breaking change" was the wrong word...

Zeitsperre: Not returning the same dimensions in an object sounds like a breaking change.

RondeauG: I think that in the end, it depends on how many people use `measures_heatmap`. If we expect the change to break the workflow of multiple people, then we should use best practices. If you are the only one really using that function, then it doesn't matter as much.

juliettelavoie: I can write a message on teams to ask if people use it.

juliettelavoie: my bet is on I'm the only one

juliettelavoie: Sarah and Marco use the ESPO-G workflow which calls this function, but if they don't have code that uses the output afterwards they should be fine.

juliettelavoie: well nobody said they used it. I think the warning might not be necessary. Can I just do the change now ?

RondeauG: Sure, go for it.

PR #171, opened 2023-03-17
Add conda recipes

### LINKED ISSUE #149: PyPI / conda implementation
### Description
For ease of use, this package should be installable and updateable through standard tools such as pip, conda, mamba.

Some tests should probably be implemented first.

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #149 

* Adds both a conda-forge compatible Anaconda build recipe, and an experimental `boa`-based recipe (used for manually generating packages).
* Updates our secrets baseline to include the SHA256 sum in the recipes.
* Excludes meta.yaml from linters, since conda-forge breaks yaml conventions.
* Updates our installation documentation.

Nope.

### DIFF
--- a/conda/xscen/meta.yaml
+++ b/conda/xscen/meta.yaml
@@ -1,0 +1,66 @@
+{% set name = "xscen" %}
+{% set version = "0.5.0" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/xscen-{{ version }}.tar.gz
+  sha256: f31df2cb52e87dd82d2fc7d340788e4edf14abccf04685a9249a2067594b721a
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  host:
+    - python >=3.8
+    - pip
+  run:
+    - python >=3.8
+    - cartopy
+    - cftime
+    - clisops
+    - dask-core
+    - flox
+    - fsspec
+    - geopandas
+    - h5netcdf
+    - intake
+    - intake-esm >=2022.9.18
+    - matplotlib-base
+    - netcdf4
+    - numpy
+    - pandas
+    - pyyaml
+    - rechunker
+    - shapely
+    - xarray
+    - xclim >=0.37
+    - xesmf >=0.7
+    - zarr
+
+test:
+  imports:
+    - xscen
+  commands:
+    - pip check
+  requires:
+    - pip
+    - pytest-json-report  # Added due to a bug in esmpy. See: https://github.com/esmf-org/esmf/issues/115
+
+about:
+  home: https://github.com/Ouranosinc/xscen
+  summary: A climate change scenario-building analysis framework, built with xclim/xarray.
+  license: Apache-2.0
+  license_file: LICENSE
+
+extra:
+  recipe-maintainers:
+    # GitHub.com
+    - Zeitsperre
+    - RondeauG
+    - aulemahal
+    - juliettelavoie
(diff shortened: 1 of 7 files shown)


conda/xscen/meta.yaml
```diff
@@ -0,0 +1,66 @@
...
+    - netcdf4
+    - numpy
+    - pandas
+    - pyyaml
+    - rechunker
+    - shapely
+    - xarray
+    - xclim >=0.37
+    - xesmf >=0.7
+    - zarr
```
RondeauG, line 43: This doesn't appear to be the same as our current `environment.yml`. Is it on purpose?

conda/xscen/recipe.yaml
```diff
@@ -0,0 +1,67 @@
...
+    - netcdf4
+    - numpy
+    - pandas
+    - pyyaml
+    - rechunker
+    - shapely
+    - xarray
+    - xclim >=0.37
+    - xesmf >=0.7
+    - zarr
```
RondeauG, line 46: Same as above

conda/xscen/recipe.yaml
```diff
@@ -0,0 +1,67 @@
...
+  home: https://github.com/Ouranosinc/xscen
+  summary: A climate change scenario-building analysis framework, built with xclim/xarray.
+  license: Apache-2.0
+  license_file: LICENSE
+
+extra:
+  recipe-maintainers:
+    # Anaconda.org
+    - Zeitsperre
+    - aule
```
RondeauG, line 67:
```suggestion
    - aulemahal
```

Zeitsperre: This is Pascal's Anaconda.org account

conda/xscen/meta.yaml
Zeitsperre: This was generated from the PyPI recipe. If this is no longer accurate, we need to update the setup requirements.

RondeauG: I see `cf_xarray>=0.7.6` in `setup.py`, for example. It should have been put here, no?

aulemahal: `cf_xarray` is never actually imported, which is why it is missing here. We might have put it in the env so that we can pin a version. It is pulled by  `xesmf` and  `xclim` in any cases.

`pygeos` was needed to accelerate some `clisops` stuff. But also never explicitly imported in xscen.

`nc-time-axis` is also needed to make plots, but `xscen` never create explicit plots.

`pyarrow` is actually explicitly needed by `xscen` (when opening catalogs), but not through an import, which might be why it is missing from pypi's recipe.

`pip` won't need to be in the env once this conda recipe exists.

RondeauG: I'm not too sure I understand. So is it fine without them here or should we change something somewhere to make them requirements?

aulemahal: Hum `cf_xarray` and `pygeos` are already in `setup.py` so I'm not sure why they are missing here...

`pyarrow` should probably be added to `setup.py`.

I'm not sure what to do with `nc-time-axis`...

RondeauG: If `nc-time-axis` is for graphs, then it should probably not be in `xscen` requirements.

Zeitsperre: I'm going to read how grayskull generates the requirements. I didn't think it did anything fancy like omitting requirements that arealready listed as requirements for other packages (I'd argue that that's an anti-feature, tbh). 

It might be based on imports found within xscen alone. If that's the case, we simply need to modify the package recipe and regenerate the binary.

Zeitsperre: I see `pygeos` in `xscen.aggregate`, so I'm not sure why it wasn't listed here. I'm also adding `cf_xarray>=0.7.6` and `pyarrow`.

RondeauG (approved):
This should be pretty good, although I'll let you or Pascal judge on the more technical parts.

PR #173, opened 2023-03-22
Fix documentation builds

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #xyz

* Adds some Makefile recipes for the documentation build steps
* Removes the llvmlite reinstallation step that was needed a few years ago to deal with some weirdness with ReadTheDocs
* Applies a consistent `package <comparison><version>` schema to environment files

No.

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -7,7 +7,7 @@ dependencies:
   # Main packages
   - cartopy
   - cftime
-  - cf_xarray >= 0.7.6
+  - cf_xarray >=0.7.6
   - clisops
   - dask
   - flox
@@ -19,7 +19,7 @@ dependencies:
   - matplotlib
   - netCDF4
   - numpy
-  - pandas!=1.5.3
+  - pandas !=1.5.3
   - pygeos
   - pyyaml
   - rechunker
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -2,13 +2,14 @@ name: xscen-dev
 channels:
   - conda-forge
 dependencies:
+  - python >=3.8
   # Don't forget to sync the changes here with environment.yml!
   # Some pins here are meant to accelerate conda and might not be related to real needs
   # Main packages
   - cartopy
   - cftime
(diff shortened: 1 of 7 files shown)


RondeauG:
I somehow can't reproduce the RtD error when running the notebook locally.

RondeauG:
Alright, I managed to reproduce the error locally. It looks like the culprit is `xarray 2023.3.0`, which gets installed during:
`python -m pip install --upgrade --upgrade-strategy eager --no-cache-dir .`

Is there a way to prevent that? Incidentally, that command also installs `pandas 1.5.3`, even though we have `pandas!=1.5.3` in `environment.yml`. Is it because it's not in `setup.py`?

RondeauG (approved):
The docs seem to be generating correctly !

environment-dev.yml
```diff
@@ -19,12 +20,12 @@ dependencies:
...
   - netCDF4
   - numpy
-  - pandas!=1.5.3
+  - pandas !=1.5.3
   - pygeos
   - pyyaml
   - rechunker
   - shapely
-  - xarray
+  - xarray <2023.3.0  # See https://github.com/Ouranosinc/xscen/pull/173
```
aulemahal, line 28: I disagree with pinning a package like this, _especially_ xarray!

The bug is in xclim, and comes from a change in a semi-private function in xarray (issue 7645 there).

Zeitsperre: Understood. How do we proceed? `xarray !=2023.3.0`?

Zeitsperre: I imagine if we can push changes to `xclim`, we can simply pin `xclim` to the next version? Will changes in `xclim` be backwards-compatible with `xarray`?

RondeauG: Done, I made an Issue.

Zeitsperre: Awesome. I'll add the changes to the changelog and update the issue to consult, then merge. Thanks!

PR #174, opened 2023-03-23
Create bounds for RotatedPole, Creep fill

### LINKED ISSUE #96: Compute bounds for curvilinear grids
### Generic Issue

Conservative methods of xESMF (and thus the SpatialAverager) need the longitude and latitude bounds to compute. This is trivial to compute on rectilinear grids: `cf_xarray` does it for `xESMF` automatically.

It is non-trivial for curvilinear grids, but possible as such a grid is the projection of a rectilinear grid (at least I think this is the definition of "curvilinear"...). Thus, it is doable by trivially generating the bounds from the original rectilinear grid and then projecting them to a PlateCarree reference system.

I once did this, using `cartopy` to parse the CF grid mappings into a `pyproj` object.  I'm not sure if `cf_xarray` would be ok with adding the code there since it would mean having these two package as dependencies (and they aren't light). However, I think we could implement something here first and move it upstream someday.

Part of my old solution lies here : https://gist.github.com/aulemahal/4873db65992369420a96834026b33470

aulemahal:
### Pull Request Checklist:
    - This PR kinda fixes #96

* Add `regrid.create_bounds_rotated_pole`. Uses `cartopy` to get `lon_bounds` and `lat_bounds` from the  Rotated Pole grid mapping. Unique to that projection, but this covers most cases at Ouranos at least... Still waiting for a proper cf-xarray - cartopy link to implement this upstream. Or to have a brilliant idea and fix the estimation in cf-xarray. In the mean time, this is good enough I think. 
    * I added checks in `regrid._regrid` and `spatial_mean` so that the conservative-rotated_pole-no_bounds case is covered automatically.

* Add `xs.spatial` for spatial tools. Currently only the `creep_fill` method to fill missing values with the mean of the neighbours. Can also be used for other purposes like making wider land-sea masks.
 
* Allow passing `GeoDataFrame` in `spatial_mean`'s `region`. This is useful if you want to open a shapefile and subset it before the mean.

No.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -15,8 +15,9 @@ New features and enhancements
 * New 'cos-lat' averaging in ``spatial_mean`` (:issue:`94`, :pull:`125`).
 * Support for computing anomalies in ``compute_deltas``  (:pull:`165`).
 * Add function ``diagnostics.measures_improvement_2d``. (:pull:`167`).
-* Add function ``regrid.create_bounds_rotated_pole`` and automatic use in ``regrid_dataset`` and ``spatial_mean``. This is temporary, while we wait for a functionning method in ``cf_xarray``.
-* Add ``spatial`` submodule with functions ``creep_weights`` and ``creep_fill`` for filling NaNs using neighbours.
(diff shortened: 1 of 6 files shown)


aulemahal:
FYI, @mccrayc @marielabonte

juliettelavoie (approved):
Just add spatial in api.rst to appear in the docs.

PR #182, opened 2023-04-04
More extract features for RDRS

### LINKED ISSUE #180: Extract Features that would be nice with RDRS
### Addressing a Problem?

In the extract step of my ESPO-G workflow, I add a few steps to handle RDRS.

1. I want to cut my region on rlat and rlon. From what I understand, clisops only cuts on lat and lon.
2. I want to add nans from my mask.

### Potential Solution

1. Add a method `sel` to region that would not be a call to clisops, just

```
arg_sel = {dim:slice(*map(float, bounds)) for dim, bounds in region['sel'].items()}
ds= ds.sel(**arg_sel )
```

2. Add a `mask` arg to add NaNs. Ideally, I would want to be able to ask for the mask in `search_data_catalogs` with `{mask:fx}` in `variables_and_freq`. Then, if `mask` is True. Apply mask on the other variables. Maybe `mask` could also be a list of variables on which to apply the mask. `mask` could also be a dataarray if the mask is not in a catalogue.
```
ds= ds.where(mask == True)
```

### Additional context

Let me know you thoughts on the best way to implement the mask.

Right now, this is all in my ESPO-G workflow, but it is not very clean... It would be better in extract.

### Contribution

juliettelavoie:
### Pull Request Checklist:
    - This PR fixes #180

* These features are in the ESPO-G workflow but I would like to have them in xscen to clean up the workflow. They were added because of RDRS.
* feature mask: Either pass a dataset of a mask or find the mask in the catalog (need to do that in search_data_catalog). The mask is used to add nans to the data.
* new region method `sel`: Instead of cutting on lat-lon (for ERA5-Land), I want to cut in rlat-rlon (for RDRS). I want this to be only defined in the config. I know it is a bit weird to put a non-clisops method in the `clisops_subset` function, but I also call this function in my workflow and I need it to cut the region regardless of if it is defined in rlat-rlon or lat-lon.  Would it be too much of a breaking change to change the name of the function ?

not, if we don't change the name of `clisops_subset`

### DIFF
--- a/xscen/extract.py
+++ b/xscen/extract.py
@@ -58,7 +58,9 @@ def clisops_subset(ds: xr.Dataset, region: dict) -> xr.Dataset:
     -----
     'region' fields:
         method: str
-            ['gridpoint', 'bbox', shape']
+            ['gridpoint', 'bbox', shape','sel']
+            If the method is `sel`, this is not a call to clisops but only a subsetting with the xarray .sel() fonction.
+            The keys are the dimension to subset and the value is turned into a slice.
         <method>: dict
             Arguments specific to the method used.
         buffer: float, optional
@@ -131,6 +133,16 @@ def clisops_subset(ds: xr.Dataset, region: dict) -> xr.Dataset:
             f" - clisops v{clisops.__version__}"
         )
 
+    elif region["method"] in ["sel"]:
+        arg_sel = {
+            dim: slice(*map(float, bounds)) for dim, bounds in region["sel"].items()
+        }
+        ds_subset = ds.sel(**arg_sel)
+        new_history = (
+            f"[{datetime.datetime.now().strftime('%Y-%m-%d %H:%M:%S')}] "
+            f"{region['method']} subsetting with arguments {arg_sel}"
+        )
+
     else:
         raise ValueError("Subsetting type not recognized")
 
@@ -157,6 +169,7 @@ def extract_dataset(
     xr_combine_kwargs: dict = None,
     preprocess: Callable = None,
     resample_methods: Optional[dict] = None,
+    mask: Union[bool, xr.Dataset] = False,
 ) -> Union[dict, xr.Dataset]:
     """Take one element of the output of `search_data_catalogs` and returns a dataset, performing conversions and resampling as needed.
 
@@ -197,6 +210,12 @@ def extract_dataset(
         If the method is not given for a variable, it is guessed from the variable name and frequency,
         using the mapping in CVs/resampling_methods.json. If the variable is not found there,
         "mean" is used by default.
+    mask: xr.Dataset, bool
+        A mask that is applied to all variables and only keeps data where it is True.
+        Where the mask is False, the variables values are replaced by NaNs.
+        The mask should have the same dimensions as the variables extracted.
+        If `mask` is a dataset, the dataset should have a variable named 'mask'.
+        If `mask` is True, it will expect a `mask` variable at xrfreq `fx` to have been extracted.
 
     Returns
     -------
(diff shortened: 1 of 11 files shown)


xscen/extract.py
```diff
@@ -197,6 +210,12 @@ def extract_dataset(
         If the method is not given for a variable, it is guessed from the variable name and frequency,
         using the mapping in CVs/resampling_methods.json. If the variable is not found there,
         "mean" is used by default.
+    mask: xr.Dataset, bool
+        A mask that is applied to all variables and only keeps data where it is True.
+        Where the mask is False, the variables values are replaced by NaNs.
```
RondeauG, line 215:
```suggestion
        Where the mask is False, variable values are replaced by NaNs.
```

xscen/extract.py
```diff
@@ -387,6 +406,23 @@ def extract_dataset(
...
+    if mask:
+        if isinstance(mask, xr.Dataset):
+            ds_mask = mask["mask"]
+        elif (
+            "fx" in out_dict and "mask" in out_dict["fx"]
+        ):  # get mask that was extracted above
+            ds_mask = out_dict["fx"]["mask"].copy()
+        else:
+            raise ValueError(
+                "No mask found. Either pass a xr.Dataset to the `mask` argument or pass a `dc` that includes a dataset with variable `mask`."
```
RondeauG, line 418:
```suggestion
                "No mask found. Either pass a xr.Dataset to the `mask` argument or pass a `dc` that includes a dataset with a variable named `mask`."
```

xscen/extract.py
```diff
@@ -387,6 +406,23 @@ def extract_dataset(
...
+            )
+
+        # iter over all xrfreq to apply the mask
+        for xrfreq, ds in out_dict.items():
+            if xrfreq != "fx":
```
RondeauG, line 423: I think we want to subset fixed fields too?

juliettelavoie: Well we for sure don't want to mask the mask and I assumed we didn't want to mask other fixed fields but if you think there is a need for it I can make it avoid just the mask.

RondeauG: I'm not 100% sure but if I ask for a mask, I'd expect it to be applied to everything?

juliettelavoie: yeah you are right

xscen/extract.py
```diff
@@ -157,6 +169,7 @@ def extract_dataset(
     xr_combine_kwargs: dict = None,
     preprocess: Callable = None,
     resample_methods: Optional[dict] = None,
+    mask: Union[bool, xr.Dataset] = False,
```
aulemahal, line 172: Could we also accept a `DataArray` ?

I don't find the "bool" option very useful, but I see that #183 would complete this feature.

aulemahal (approved):
I agree that `clisops_subset` is a bit misnamed. I would support a renaming, and may be a move to `xscen.spatial` ?

We could lower the breakingness of this change by keeping a `clisops_wrapper` that emits a FutureWarning for one version.

xscen/extract.py
juliettelavoie: sure for DataArray.
I guess it is only useful for me now that I already have a mask that is in the catalogue.

RondeauG:
I just saw that `clisops_subset` still exists as is. I think that we should rather have:
```
def clisops_subset(ds, region):
     warning(blabla)
     ds_subset = xs.spatial.subset(ds, region)
     return ds_subset
```

So as not to duplicate the code twice and have an outdated version, especially since the new function should work perfectly well with the old arguments.

juliettelavoie:
> I just saw that `clisops_subset` still exists as is. I think that we should rather have:
> 
> ```
> def clisops_subset(ds, region):
>      warning(blabla)
>      ds_subset = xs.spatial.subset(ds, region)
>      return ds_subset
> ```
> 
> So as not to duplicate the code twice and have an outdated version, especially since the new function should work perfectly well with the old arguments.

done!

PR #184, opened 2023-04-05
Refactor 'region'

### LINKED ISSUE #99: Docstring changes for clarity
A few tripfalls that came up during the training session (for me, anyway):
 - `xscen.ProjectCatalog` - It should be clearer in the docstring that the ProjectCatalog needs to be initialized empty, then populated.
 - `xscen.extract_dataset`:
```
Notes
-----
'region' fields:
    name: str
        Region name used to overwrite domain in the catalog.
    method: str
        ['gridpoint', 'bbox', shape']
    <method>: dict -> This should become `method_kwargs`
        Arguments specific to the method used.
    buffer: float, optional
        Multiplier to apply to the model resolution.
```
Additionally, it should be made clear in the logs that `buffer` is dropped for subset_gridpoint (at the `INFO` level at least). It would also be good to have a warning that `buffer` in xscen is *not* treated the same as `buffer` in CLISOPS.

### LINKED ISSUE #101: Rework clisops_subset arguments
### Generic Issue

* xscen version: 0.4.0

### Description
`xscen.extract.clisops_subset` is currently defined that way: `clisops_subset(ds: xr.Dataset, region: dict)`. 

This means that:

- `region` has to be an awkward dictionary, where the required entries aren't too obvious.
- The kwargs have to be included in `region`. This wouldn't be too bad (but still not great), but the key that has to be used is really not obvious either.

### Suggestion

1. Rework the arguments to be more explicit: `clisops_subset(ds: xr.Dataset, name: str, buffer: float, method: str, **kwargs)`
2. Where clisops_subset is called, such as `extract_dataset`, use `region` still, but reworked in a way where we can use `clisops_subset(ds: xr.Dataset, **region)`

For example:
```
region = {
     name: example
     method: shape
     buffer: 1.5
     shape: {
          shape: path/to/file.shp
     }
}
```
would become:
```
region = {
     name: example
     method: shape
     buffer: 1.5
     shape: path/to/file.shp
}
```

The one issue we'd have to solve is `buffer`, which is currently in an awkward place anyway since it behaves differently depending on the method used (#99).

RondeauG:
### Pull Request Checklist:
    - This PR fixes #99
    - This PR fixes #101

* Refactors `region` to make it more easily understandable and make use of `**kwargs` instead of the awkward formatting that it currently has.
* Makes the arguments of `xs.spatial.subset` more explicit. `region` is still used in `xs.extract_dataset` and `xs.spatial_mean`, but not in `xs.spatial.subset`.

* Multiple things get deprecated due to the changes in formatting, but everything should still be supported for the time being.


HISTORY.rst
```diff
@@ -31,6 +32,8 @@ Breaking changes
 * The 'datasets' dimension of the output of ``diagnostics.measures_heatmap`` is renamed 'realization'. (:pull:`167`).
 * `_subset_file_coverage` was renamed `subset_file_coverage` and moved to ``catalog.py`` to prevent circular imports. (:pull:`170`).
 * `extract_dataset` doesn't fail when a variable is in the dataset, but not `variables_and_freqs`. (:pull:`185`).
+* ``xscen.extract.clisops_subset`` has been deprecated and replaced by ``xscen.spatial.subset``. That new function no longer uses `region` as an argument. (:pull:`182`, :pull:`184`).
+* The argument `region`, used in multiple functions, has been slightly reformatted. (:issue:`99`, :issue:`101`, :pull:`184`).
```
aulemahal, line 36:
```suggestion
```
Not so important, but seems to me that the announcements above are enough ?

xscen/aggregate.py
```diff
@@ -294,6 +294,7 @@ def spatial_mean(
     ds: xr.Dataset,
     method: str,
     *,
+    spatial_subset: bool = False,
     call_clisops: bool = False,
     region: dict = None,
```
aulemahal, line 299:
```suggestion
    spatial_subset: bool = None,
    call_clisops: bool = False,
    region: dict = None,
```
Suggestion to do the following:

If `spatial_subset` is None, `region` is not None and `method` is in `['cos-lat', 'mean'], then `spatial_subset=True`.

For methods "xESMF" and "interp_centroid", the information of `region` is used for the spatial mean : subsetting is implicit. But with the two other methods there is an ambiguity.  I just had this problem where I assumed that `spatial_mean` was taking the average over my given region, but wasn't passing `call_clisops=True` so it was taking the average over all the domain. It doesn't make sense to me to accept a `region` arg and just ignore it silently like this.

HISTORY.rst
RondeauG: Good enough. I wasn't sure if they needed to be repeated or not.

xscen/aggregate.py
RondeauG: I fear that dynamically changing the behaviour of `None` might be confusing? An alternative that would have worked in your example would be to raise an error if those 3 conditions are met.

aulemahal: On the contrary, I feel that a default of None is much more like a "unset" which means the default behaviour depends on something else. 
In our case here, as discussed earlier, we could have chosen to always subset, but that _might_ introduce some bugs in some edge cases (subsetting without buffer before averaging with xesmf, for example).

xscen/aggregate.py
```diff
@@ -340,11 +341,11 @@ def spatial_mean(
     -----
     'region' required fields:
```
juliettelavoie, line 342: It's not a required field, but I would still add name here.

juliettelavoie (approved):
I agree with Pascal's suggestion on changing the behavior of spatial_subset in the presence of region.

xscen/aggregate.py
RondeauG: When you have time, could you check the new commit to make sure that my implementation is what you had in mind?

aulemahal (approved):

PR #187, opened 2023-04-13
Template for unit tests

RondeauG:
### Pull Request Checklist:
    - Not really (at least not enough to resolve the issue)

* Implements a working template on which to start creating unit tests for xscen.
* Added NotImplementedError in `climatological_mean` and `compute_deltas` when using daily data.

* No.

### Other information:

This is a start for #9

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -3,9 +3,7 @@ import shutil
 from functools import partial
 from pathlib import Path
 
-import pandas as pd
 import pytest
-import xarray as xr
 from xclim.testing.helpers import test_timeseries
 
 notebooks = Path().cwd().parent / "docs" / "notebooks"
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2,7 +2,9 @@
 import shutil
 from pathlib import Path
 
+import pandas as pd
 import pytest
+import xarray as xr
 
 notebooks = Path().cwd().parent / "docs" / "notebooks"
 
@@ -20,3 +22,29 @@ def cleanup_notebook_data_folder(request):
             shutil.rmtree(data)
 
     request.addfinalizer(remove_data_folder)
+
+
+@pytest.fixture
+def tas_series_1d():
+    """Create a temperature time series."""
+
+    def _tas_series(values, start="7/1/2000", units="K", freq="D", as_dataset=True):
+        coords = pd.date_range(start, periods=len(values), freq=freq)
+        da = xr.DataArray(
+            values,
+            coords=[coords],
+            dims="time",
+            name="tas",
+            attrs={
+                "standard_name": "air_temperature",
+                "cell_methods": "time: mean within days",
+                "units": units,
+            },
+        )
+        da.name = "tas"
+        if as_dataset:
+            return da.to_dataset()
+        else:
+            return da
+
+    return _tas_series
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,10 +1,8 @@
 # noqa: D100
 import shutil
-from functools import partial
 from pathlib import Path
 
 import pytest
-from xclim.testing.helpers import test_timeseries
 
 notebooks = Path().cwd().parent / "docs" / "notebooks"
 
@@ -22,10 +20,3 @@ def cleanup_notebook_data_folder(request):
             shutil.rmtree(data)
 
     request.addfinalizer(remove_data_folder)
-
-
-@pytest.fixture
-def tas_series():
-    """Return mean temperature time series."""
-    _tas_series = partial(test_timeseries, variable="tas")
-    return _tas_series
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,10 +1,12 @@
 # noqa: D100
 import shutil
+from functools import partial
 from pathlib import Path
 
 import pandas as pd
 import pytest
 import xarray as xr
+from xclim.testing.helpers import test_timeseries
 
 notebooks = Path().cwd().parent / "docs" / "notebooks"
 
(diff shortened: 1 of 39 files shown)


RondeauG:
@aulemahal @Zeitsperre Feel free to work directly on this branch and to change/remove what I've done. Once we're happy with the formatting, we'll merge it and I'll start making real unit tests.

tests/conftest.py
```diff
@@ -20,3 +22,29 @@ def remove_data_folder():
...
+                "units": units,
+            },
+        )
+        da.name = "tas"
+        if as_dataset:
+            return da.to_dataset()
+        else:
+            return da
+
+    return _tas_series
```
Zeitsperre, line 50: I think we should explore the xclim-based solution. Would you want me and/or @aulemahal to add something there first?

RondeauG: Yeah, this is just what I had at the time of our discussion, and helps to show the modifications I had to do (such as expose `freq`). We should indeed work on modifying `xclim` first.

RondeauG: @Zeitsperre @aulemahal 
This new commit will fail because `test_timeseries` isn't yet accessible outside of `xclim`'s master branch, but is this closer to what we had in mind ? Or did we want to remove fixtures altogether ?

tests/conftest.py
```diff
@@ -20,3 +24,10 @@ def remove_data_folder():
             shutil.rmtree(data)
 
     request.addfinalizer(remove_data_folder)
+
+
+@pytest.fixture
+def tas_series():
+    """Return mean temperature time series."""
+    _tas_series = partial(test_timeseries, variable="tas")
+    return _tas_series
```
aulemahal, line 33: This is the xclim solution, where we avoided a huge PR by keeping the same fixture. However, I'm not sure I see any gain in using fixtures here ?

I would have suggested importing `test_timeseries` in each test file and using the function directly in the tests. This will  make for a simpler `conftest.py`.

RondeauG:
```
import numpy as np
import xscen as xs
from xclim.testing.helpers import test_timeseries
import pytest


class TestClimatologicalMean:
    def test_daily(self):
        ds = test_timeseries(np.tile(np.arange(1, 13), 3), variable="tas", start="2001-01-01", freq="D", as_dataset=True)
        with pytest.raises(NotImplementedError):
            xs.climatological_mean(ds)
```
I get this error when trying to use it directly:
```
___________________________________________________________________________________________________________________ ERROR at setup of test_timeseries ____________________________________________________________________________________________________________________
  def test_timeseries(
E       fixture 'values' not found
>       available fixtures: anyio_backend, anyio_backend_name, anyio_backend_options, cache, capfd, capfdbinary, caplog, capsys, capsysbinary, cleanup_notebook_data_folder, cov, doctest_namespace, include_metadata_in_junit_xml, json_metadata, metadata, monkeypatch, no_cover, pytestconfig, record_property, record_testsuite_property, record_xml_attribute, recwarn, tas_series, tmp_path, tmp_path_factory, tmpdir, tmpdir_factory, xdoctest_namespace
>       use 'pytest --fixtures [testpath]' for help on them.

```

aulemahal: Je'l savais merde.

```
from xclim.testing.helpers import test_timeseries as timeseries
```

When a member function of a `test.py` file starts with `test_` it is considered a test and pytest tries to run it so. `values` is the first arg of the function and pytest assumes it's a fixture.

RondeauG:
What I'm uploading here is a version that uses `test_timeseries` and hits pretty much all of `climatological_mean`. Since this is the best time to establish a format, I'm open to all suggestions!

tests/test_aggregate.py
```diff
@@ -138,6 +138,10 @@ def test_calendars(self):
             as_dataset=True,
         )
 
+        assert isinstance(
+            xs.climatological_mean(convert_calendar(ds, "standard")),
+            cftime.Datetime360Day,
+        )
```
aulemahal: what ? Standard -> Datetime360Day ?

My commit changed the logic here anyway.

RondeauG: I think that's some bad copy-paste 😄

aulemahal:
@RondeauG 
Please see the changes introduced in my latest commit. 

- Instead of looping inside the test function, I used `pytest.mark.parametrize`. The benefit here is that if a single element fails:
    - The others are also run
    - The pytest error output specifies the exact element causing the bug
    - And we can run a single element by hand if needed, while debugging.
- I switched to conventionnal `assert` when we are comparing scalars.
- I rewrote a check with a `set` logic instead of an `all` of a in comprehension list.
- Rewrote the calendar check to use only xarray properties and pytest parametrize

RondeauG:
This should be good for review.

Zeitsperre (approved):
Will wait on @aulemahal's approval as well

xscen/aggregate.py
```diff
@@ -174,9 +187,7 @@ def climatological_mean(
             else new_history
         )
         ds_rolling[vv].attrs["history"] = history
-
-    if to_level is not None:
-        ds_rolling.attrs["cat:processing_level"] = to_level
+    ds_rolling.attrs["cat:processing_level"] = to_level
```
aulemahal, line 190: This doesn't fit the docstring, which says None is possible!
I'm guessing you want to change the docstring?

aulemahal (approved):
One comment to the changes in `climatological_mean`.

For the test, I think this is a good example of what we want. Even, it is on the exhaustive side, which is good.

Warning : all tests here are done on artificial data, and thus it may fail to tests some real-life caveats. When adding more tests, we could make use of xclim's test data too.

xscen/aggregate.py
RondeauG: Ah, true! I got confused because the default is a string, not `None`. I'll reverse that.

PR #191, opened 2023-04-25
YAML notebook

### LINKED ISSUE #8: Make the YAML-centric notebook
### Description

The Notebook on YAML usage still needs to be written. My idea here was to copy `Getting Started`, but using a config file instead of explicitely calling the arguments.

RondeauG:
### Pull Request Checklist:
    - This PR fixes #8

* Added a YAML-centered notebook
* Added `@parse_config` to a few functions that were missing the wrapper.
* Fixed the formatting of alert boxes in the documentation.

- No


docs/index.rst
```diff
@@ -33,6 +33,7 @@ Features
     notebooks/3_diagnostics
     notebooks/4_ensemble_reduction
     notebooks/5_warminglevels
+    notebooks/6_yaml
```
aulemahal, line 36: I would suggest the "6_config"  name instead ? To me "yaml" is an implementation detail.

aulemahal (commented):
Few comments.

I think hardcoding a list of functions like that is prone to error, I will push a dynamic option in a few seconds.

I think we also could  show an example of `CONFIG.set`, to set values in the workflow itself.
And have a word about the "special" auto-configuring sections ? ('logging', 'xclim', 'xarray', 'warning').

Also, here would be a good place for an example using `argparse` and passing args to `load_config`. I could contribute that part.

docs/index.rst
RondeauG: Done

RondeauG:
> Few comments.
> 
> I think hardcoding a list of functions like that is prone to error, I will push a dynamic option in a few seconds.
> 
> I think we also could show an example of `CONFIG.set`, to set values in the workflow itself. And have a word about the "special" auto-configuring sections ? ('logging', 'xclim', 'xarray', 'warning').
> 
> Also, here would be a good place for an example using `argparse` and passing args to `load_config`. I could contribute that part.

Would you be able to contribute those? I'm not really familiar with those methods.

xscen/spatial.py
```diff
@@ -119,6 +123,7 @@ def _dot(arr, wei):
     )
 
 
+@parse_config
```
juliettelavoie, line 126: I'm not sure it's a good idea to add parse_config to this function because it is called by other functions that might not use it in the same way.
Somebody might use `subset` in one way in their workflow and need to use it in a different way in `spatial_mean` , but not realise that the config is used in both cases...

juliettelavoie (approved):
looks good! 
Not sure I get the point of the last cell of the notebook though...

aulemahal (approved):
I approve my additions! ;)

PR #192, opened 2023-04-25
Standardize 'period' / 'periods'

### LINKED ISSUE #87: More intuitive formatting for 'periods'
### Addressing a Problem?

`periods` is used across multiple functions in `xscen`, but it is not always easy to know which format is accepted or not. Sometimes `[start, end]` will work, sometimes it needs to be a list of lists `[[start, end]]`. Sometimes `int` works, sometimes `str` are required... It's honestly quite a mess that we should improve.

### Potential Solution

Support multiple formats (list & list of lists, int & str) for all functions that have a `periods` argument.

### Additional context

_No response_

### Contribution

RondeauG:
### Pull Request Checklist:
    - This PR fixes #87

* Slightly going back from #87, `period` is always a list, while `periods` is more flexible.
* A new function `utils.standardize_periods` to always have the same formatting.
* Added a `coverage_kwargs` to `search_data_catalogs` to add a way to modify the default arguments of `subset_file_coverage`

* The one breaking change is the renaming of `simulation_period`. Since it is both positional and required, I don't see a way to safely replace it by `simulation_periods`.


xscen/extract.py
```diff
@@ -532,7 +533,9 @@ def search_data_catalogs(
     match_hist_and_fut: bool, optional
         If True, historical and future simulations will be combined into the same line, and search results lacking one of them will be rejected.
     periods : list
-        [start, end] of the period to be evaluated (or a list of lists).
+        Either [start, end] or list of [start, end] for the periods to be evaluated.
+    coverage_kwargs : dict
+        If required, arguments to pass to subset_file_coverage (only used when periods is not None).
```
juliettelavoie, line 538: what does "if required" mean ?

xscen/biasadjust.py
```diff
@@ -187,7 +189,7 @@ def train(
 def adjust(
     dtrain: xr.Dataset,
     dsim: xr.Dataset,
-    simulation_period: list,
+    simulation_periods: list,
```
juliettelavoie, line 192: I don't have an idea on how to not have a breaking change, but I might have an idea on breaking it more :P
Should we juste call it `periods` like everywhere else? same for `reference_period` -> `period` in `train`.

juliettelavoie (approved):

xscen/biasadjust.py
```diff
@@ -267,12 +269,9 @@ def adjust(
...
     # do the adjustment for all the simulation_period lists
-    if isinstance(
-        simulation_period[0], str
-    ):  # if only one period, turn it into a list of list
-        simulation_period = [simulation_period]
+    simulation_periods = standardize_periods(simulation_periods)
     slices = []
-    for period in simulation_period:
+    for period in simulation_periods:
         sim = sim.sel(time=slice(period[0], period[1]))
```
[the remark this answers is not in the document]
RondeauG: Oof, indeed!

xscen/biasadjust.py
RondeauG: Those two are indeed the only instances that are not `period(s)` (and `tas_baseline_period` in `subset_warming_level`, but that's different enough). Since we're gonna break it anyway, I'm not against renaming them.

@aulemahal Thoughts?

aulemahal: Good with me!

PR #214, opened 2023-06-20
add scripting utils

juliettelavoie:
### Pull Request Checklist:
    - This PR fixes #xyz

* `save_and_update`: nothing new, just putting things that exists all together so that I only have one line at the end of each tasks.
** I removed the "move" part of that function, because we want to encourage Ouranos users to write everything in one place and than use `move_and_delete`.
* `move_and_delete`: In a workflow, when a simulation is done, you can save what you want to keep (final, maybe regridded and log file) and delete the workdir.
* added catutils to the API for the docs

no

### Other information:
* how do we do tests for io stuff ?
* I am thinking of using these function in template 1, but I am not sure. It might be better to see that save and update are different functions when you are learning xscen? Is `move_and_delete` too advanced for basic workflow ?

### DIFF
--- a/xscen/scripting.py
+++ b/xscen/scripting.py
@@ -443,7 +443,7 @@ def move_and_delete(moving, pcat, deleting):
             if Path(source).exists():
                 logger.info(f"Moving {source} to {dest}.")
                 sh.move(source, dest)
-                if Path(dest).suffix in ["zarr", "nc"]:
+                if Path(dest).suffix in [".zarr", ".nc"]:
                     ds = xr.open_dataset(dest)
                     pcat.update_from_ds(ds=ds, path=dest)
             else:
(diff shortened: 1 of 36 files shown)


xscen/scripting.py
```diff
@@ -338,3 +344,119 @@ def skippable(seconds: int = 2, task: str = "", logger: logging.Logger = None):
...
+    if file_format is None:
+        file_format = build_path_kwargs.get("format", None)
+        if path is not None and "." in path:
+            file_format = Path(path).suffix.split(".")[-1]
+        else:
+            file_format = ds.attrs.get("cat:format", "zarr")
+
+    # get path
+    if path is not None:
+        path = path.format(**get_cat_attrs(ds))  # fill path with attrs
```
aulemahal, line 398:
```suggestion
        path = str(path).format(**get_cat_attrs(ds))  # fill path with attrs
```
Since we accept Path objects too.

aulemahal (approved):
Looks good.

I'm also not sure who this can be integrated in the test suite. You could generate a file and move it around, checking that the catalog is updated. It would be a very shallow test...

PR #222, opened 2023-07-07
Dates as datetime64[ms] - remove `driving_institution`

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

* The `date_start` and  `date_end` columns are casted with a `datetime64[ms]` dtype (not a Period)
* Improvements to `date_parser`.
* Rewrite of `subset_file_coverage`.
* Removal of `driving_institution` as an official xscen column.
* pin of pandas >= 2

Pandas 2 now supports datetime columns with a s, ms and us resolution, instead of the old ns default. This allows storing dates from before 1677 and after 2242. However, this support is still partial as many of the datetime manipulation methods will still fail on "out of bounds" dates. This includes: `pd.read_csv` and `pd.to_datetime`... Because of this bug, I had to implement the parsing directly in the `DataCatalog`'s init, using a solution proposed [on stackoverflow](https://stackoverflow.com/a/76613404/9291575).

Even with this strange workaround, opening `simulation.json` went from 3 s to 800 ms on my machine !

The change had repercussions in other parts of xscen, especially `date_parser` and `subset_file_coverage`. I adapted the former to output `pd.Timestamp` objects by default and the latter to use more of the `Interval` magic pandas can already do with datetime bounds.

I also used this PR to remove `driving_institution` from the official columns, as discussed.

The default output of `date_parser` has changed.

The default dtype of `date_start` and `date_end` has changed.

The `driving_institution` column has been removed.

### Other information:
This required pinning pandas >= 2, clisops >= 0.10. The latter pin allowed unpinning python.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -46,7 +46,7 @@ jobs:
 #            tox-build: "py39"
           - python-version: "3.10"
 #            tox-build: "py310"
-#          - python-version: "3.11"
+          - python-version: "3.11"
 #            tox-build: "py311"
     defaults:
       run:
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -8,7 +8,7 @@ Contributors to this version: Gabriel Rondeau-Genesse (:user:`RondeauG`), Pascal
 
 Announcements
 ^^^^^^^^^^^^^
-* Dropped support for Python 3.8. (:pull:`199`).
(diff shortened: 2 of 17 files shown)


aulemahal:
Feel free to push changes and merge this whenever you may want to. I'll be on vacation for the next two weeks.

RondeauG (approved):
LGTM!

PR #227, opened 2023-07-24
Fix subset file coverage

### LINKED ISSUE #226: Incorrect behaviour of subset_file_coverage when multiple files are in range
### Setup Information

- xscen version: 0.6.12

### Description

Since the recent change, `catalog.subset_file_coverage` fails when multiple files are located within the interval, because `guessed_length` results in multiple values instead of just one:

```
guessed_length = pd.IntervalIndex.from_arrays(
                intervals[files_in_range].map(
                    lambda x: max(x.left, period_interval.left)
                ),
                intervals[files_in_range].map(
                    lambda x: min(x.right, period_interval.right)
                ),
            ).length
```

Which gives:
`ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()`

### Steps To Reproduce

Search for multiple years of CRCM5 data (which is saved monthly).

### Additional context

Applying `sum(guessed_length, datetime.timedelta())` will make the line `guessed_length / period_length < coverage` work as intended.

Found by @vindelico 

### Contribution

aulemahal:
### Pull Request Checklist:
    - This PR fixes #226 
- [ ] HISTORY.rst has been updated (with summary of main changes).

* Fix the usage of `IntervalIndex.length`.
* Cast period bounds to string before calling `date_parser`. This is not the most elegant method, I'll suggest something better tomorrow.

No.

### DIFF
--- a/xscen/catalog.py
+++ b/xscen/catalog.py
@@ -935,8 +935,8 @@ def subset_file_coverage(
     files_to_keep = []
     for period in periods:
         period_interval = pd.Interval(
-            date_parser(period[0]),
-            date_parser(period[1], end_of_period=True),
+            date_parser(str(period[0])),
+            date_parser(str(period[1]), end_of_period=True),
             closed="both",
         )
         files_in_range = intervals.overlaps(period_interval)
@@ -960,7 +960,7 @@ def subset_file_coverage(
                 intervals[files_in_range].map(
                     lambda x: min(x.right, period_interval.right)
                 ),
-            ).length
+            ).length.sum()
 
             if guessed_length / period_length < coverage:
                 logging.warning(
--- a/xscen/catalog.py
+++ b/xscen/catalog.py
@@ -935,8 +935,8 @@ def subset_file_coverage(
     files_to_keep = []
     for period in periods:
         period_interval = pd.Interval(
-            date_parser(str(period[0])),
-            date_parser(str(period[1]), end_of_period=True),
+            date_parser(period[0]),
+            date_parser(period[1], end_of_period=True),
             closed="both",
         )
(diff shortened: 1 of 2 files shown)


xscen/catalog.py
```diff
@@ -935,8 +935,8 @@ def subset_file_coverage(
     files_to_keep = []
     for period in periods:
         period_interval = pd.Interval(
-            date_parser(period[0]),
-            date_parser(period[1], end_of_period=True),
+            date_parser(str(period[0])),
```
RondeauG, line 938: `standardize_periods` is called earlier in that function, so I don't think this is needed?

RondeauG (approved):
I published the Issue yesterday evening, to fix it this morning... You're way too fast :smile:

xscen/catalog.py
```diff
@@ -960,7 +960,7 @@ def subset_file_coverage(
                 intervals[files_in_range].map(
                     lambda x: min(x.right, period_interval.right)
                 ),
-            ).length
+            ).length.sum()
```
RondeauG, line 963: Have you tested it? I don't see a `.sum()` in the list of functions.

xscen/catalog.py
aulemahal: Ah woups indeed. Hum, why did I need that in my local tests then? I'll investigate.

xscen/catalog.py
RondeauG: Ah indeed, I think that the output type of `IntervalIndex.length` is just not documented: https://pandas.pydata.org/docs/reference/api/pandas.IntervalIndex.length.html

PR #237, opened 2023-08-11
Stricter build_path / better end_of_period

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

Two different things, but that I both realized when updating the catalogs.

#### Stricter build_path
When I migrated from miranda, I relaxed the "structure" code because it felt too restricting and I wanted to simplify the logic. However, this wasn't a good idea. When moving the ESPO-R5-E5L indicators, I forgot to include the "experiment" field somewhere and used build_path to copy the files. Result : one scenario overwrote the other, I lost half of the data. 

This PR changes the things a bit, the main change being :  All facets, except those marked optional, are necessary. `build_path` will FAIL if any is missing.

And:
- New way to specific a folder level in the schema : "(<facet name>)" (with parenthesis). This facet is marked as optional and if it is missing from the data, the level is skipped.
- Removal of the "option: " structure from the folder schema. This was only used to put "[bias_adjust_project version]" if the former was non-null. Instead, the schemas are duplicated : one for the "raw" case and one for the "biasadjusted" case. (And similarly for derived data).
- The previous point allowed me to rewrite `_get_needed_fields` without the funcky magic needed before.
- Removal of the "strict" keyword. It is always strict. The previous `strict=True` was overly strict because of caveats of `_get_needed_fields`. Those are now fixed and `strict=False` shouldn't be needed.
- Some syntax in the yml file changed.
- Passing a dataframe/catalog will now also add a "new_path_type" column to the output, so one can make sure all entries have been constructed from the same schema.

#### Better end_of_period
When I updated pandas to 2, I modified `date_parser` and it changed how the "end_of_period" was handled.

```python
date_parser('2020', end_of_period=True)
# Initial xscen: "2020-12-31 23:00"
# Current xscen : "2020-12-31 00:00:00"
# This PR: "2020-12-31 23:59:59"
```
Thus, when searching for a coverage, the error due to the hour of the period end will be reduced.

`build_path` is now always strict.

### Other information:
Do you agree?

I could re-implement `strict=False` if needed. It would mark all fields as optional on-the-fly. The default would stay "True".

### DIFF
--- a/xscen/data/file_schema.yml
+++ b/xscen/data/file_schema.yml
@@ -124,7 +124,7 @@ raw-hydro-sims-ba:
     - facet: type
       value: simulation-hydro
     - facet: processing_level
-      value: [ rbiasadjusted ]
+      value: [ biasadjusted ]
   folders:
     - type
     - hydrology_project
--- a/xscen/data/file_schema.yml
+++ b/xscen/data/file_schema.yml
@@ -266,4 +266,4 @@ derived-hydro-reconstruction:
     - processing_level
     - xrfreq
     - variable
-  filename: [ variable, xrfreq, hydrology_project, hydrology_institution, hydrology_source, institution, source, version, member, domain, processing_level, DATES ]
+  filename: [ variable, xrfreq, hydrology_project, hydrology_source, hydrology_member, institution, source, version, member, domain, processing_level, DATES ]
--- a/xscen/data/file_schema.yml
+++ b/xscen/data/file_schema.yml
@@ -18,11 +18,11 @@
 #     filename: # The file name schema, a list of facet names. If a facet is empty, it will be skipped. Elements will be separated by "_".
 #               # The special "DATES" facet will be replaced by the most concise way found to define the temporal range covered by the file.
 ---
-### Raw data
+### Original / raw data
 #
-# These schemas are meant for raw-like data, usually hourly or daily, used as input for scenario projects.
-# It includes bias-adjusted datasets.
-raw-non-sims:
+# These schemas are meant for primary data, usually hourly or daily, used as input for scenario projects.
+# It usually includes datasets with processing_level "raw" or "biasadjusted
+original-non-sims:
   with:
     - facet: type
       value: [station-obs, reconstruction, forecast]
@@ -37,7 +37,7 @@ raw-non-sims:
     - frequency
     - variable
   filename: [ variable, frequency, domain, institution, source, version, member, DATES ]
-raw-sims-raw:
+original-sims-raw:
   with:
     - facet: type
       value: simulation
@@ -57,7 +57,7 @@ raw-sims-raw:
     - frequency
     - variable
   filename: [ variable, frequency, bias_adjust_project, version, mip_era, activity, domain, institution, source, driving_model, experiment, member, DATES ]
-raw-sims-ba:
+original-sims-ba:
   with:
     - facet: type
       value: simulation
@@ -78,7 +78,7 @@ raw-sims-ba:
     - frequency
     - variable
   filename: [ variable, frequency, bias_adjust_project, version, mip_era, activity, domain, institution, source, driving_model, experiment, member, DATES ]
-raw-hydro-reconstruction:
+original-hydro-reconstruction:
   with:
     - facet: type
       value: reconstruction-hydro
@@ -88,15 +88,15 @@ raw-hydro-reconstruction:
     - type
     - domain
     - hydrology_project
-    - hydrology_institution
     - hydrology_source
+    - (hydrology_member)
     - institution
     - [source, version]
     - (member)
     - frequency
     - variable
   filename: [ variable, frequency, domain, hydrology_project, hydrology_institution, hydrology_source, institution, source, version, member, DATES ]
-raw-hydro-sims-raw:
+original-hydro-sims-raw:
   with:
     - facet: type
       value: simulation-hydro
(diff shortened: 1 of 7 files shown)


xscen/data/file_schema.yml
```diff
@@ -26,46 +24,65 @@
...
     - domain
     - institution
     - [source, version]
-    - member
+    - (member)
     - frequency
     - variable
   filename: [ variable, frequency, domain, institution, source, version, member, DATES ]
-raw-sims:
+raw-sims-raw:
```
juliettelavoie, line 40: we use raw in two different contexts:
1) raw vs bias adjusted
2) raw vs derived

Can we find a different word for 2 ? I think it would be less confusing and avoid "raw-sims-raw".
a few suggestions: `direct`, `basic`, `original`,  `primary`, `elemental` ?

RondeauG: I prefer `original` from this list.

xscen/data/file_schema.yml
```diff
@@ -75,35 +92,58 @@ raw-hydro-reconstruction:
...
+  with:
+    - facet: type
+      value: simulation-hydro
+    - facet: processing_level
+      value: [ raw ]
+  folders:
+    - type
+    - hydrology_project
+    - hydrology_institution
+    - hydrology_source
```
RondeauG, line 109: `hydrology_member` is missing from this list. Also, after reflection, I think that similar to how `bias_adjust_institution` is in the catalog but not the structure, we could remove `hydrology_institution`. So:

```
    - hydrology_project
    - hydrology_source
    - (hydrology_member)
```

xscen/data/file_schema.yml
```diff
@@ -5,105 +5,145 @@
...
-    - hydrology_institution
     - hydrology_source
+    - (hydrology_member)
     - institution
     - [source, version]
-    - member
+    - (member)
     - frequency
     - variable
   filename: [ variable, frequency, domain, hydrology_project, hydrology_institution, hydrology_source, institution, source, version, member, DATES ]
```
RondeauG, line 98:
```suggestion
  filename: [ variable, frequency, domain, hydrology_project, hydrology_source, hydrology_member, institution, source, version, member, DATES ]
```

xscen/data/file_schema.yml
```diff
@@ -5,105 +5,145 @@
...
+    - activity
+    - domain
+    - institution
+    - source
+    - (driving_model)
+    - experiment
+    - member
+    - frequency
+    - variable
+  filename: [ variable, frequency, hydrology_project, hydrology_institution, hydrology_source,  bias_adjust_project, version, mip_era, activity, domain, institution, source, driving_model, experiment, member, DATES ]
```
RondeauG, line 121:
```suggestion
  filename: [ variable, frequency, hydrology_project, hydrology_source, hydrology_member, bias_adjust_project, version, mip_era, activity, domain, institution, source, driving_model, experiment, member, DATES ]
```

xscen/data/file_schema.yml
```diff
@@ -5,105 +5,145 @@
...
     - domain
     - institution
     - source
-    - driving_model
+    - (driving_model)
     - experiment
     - member
     - frequency
     - variable
   filename: [ variable, frequency, hydrology_project, hydrology_institution, hydrology_source,  bias_adjust_project, version, mip_era, activity, domain, institution, source, driving_model, experiment, member, DATES ]
```
RondeauG, line 145:
```suggestion
  filename: [ variable, frequency, hydrology_project, hydrology_source, hydrology_member, bias_adjust_project, version, mip_era, activity, domain, institution, source, driving_model, experiment, member, DATES ]
```

xscen/data/file_schema.yml
```diff
@@ -136,36 +193,56 @@ derived-sims:
...
+    - institution
+    - source
+    - (driving_model)
+    - experiment
+    - member
+    - domain
+    - processing_level
+    - xrfreq
+    - variable
+  filename: [ variable, xrfreq, hydrology_project, hydrology_institution, hydrology_source, bias_adjust_project, version, mip_era, activity, institution, source, driving_model, experiment, member, domain, processing_level, DATES ]
```
RondeauG, line 230:
```suggestion
  filename: [ variable, xrfreq, hydrology_project, hydrology_source, hydrology_member, bias_adjust_project, version, mip_era, activity, institution, source, driving_model, experiment, member, domain, processing_level, DATES ]
```

xscen/data/file_schema.yml
```diff
@@ -175,16 +252,16 @@ derived-hydro-sims:
   filename: [ variable, xrfreq, hydrology_project, hydrology_institution, hydrology_source, bias_adjust_project, version, mip_era, activity, institution, source, driving_model, experiment, member, domain, processing_level, DATES ]
```
RondeauG, line 252:
```suggestion
  filename: [ variable, xrfreq, hydrology_project, hydrology_source, hydrology_member, bias_adjust_project, version, mip_era, activity, institution, source, driving_model, experiment, member, domain, processing_level, DATES ]
```

xscen/data/file_schema.yml
```diff
@@ -175,16 +252,16 @@ derived-hydro-sims:
...
-    - hydrology_institution
     - hydrology_source
+    - (hydrology_member)
     - institution
     - [source, version]
-    - member
+    - (member)
     - domain
     - processing_level
     - xrfreq
```
RondeauG, line 267: There's also a filename to fix below this line, but I don't have access to it from the PR..

PR #243, opened 2023-08-23
Prep 0.7.1 - update deps

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

Update dependencies : 
 
* Add missing deps from the conda-forge comments
* Pin intake-esm >= 2023.07.07 to remove pin on pydantic.
* Update python's pin 
* Remove pygeos and pin shapely > 2 to get the same perf.

I also removed the `recipe.yaml` file because I thought that it was only a repetition of the `meta.yaml` ?

No


conda/xscen/recipe.yaml
Zeitsperre, line 1: I hate to be that guy, but this is a bit different from meta.yaml (it's for a tool called `boa` which is a much nicer way of building packages locally). Would it be possible to apply the same changes as meta.yaml here for now?

The conda folder is completely obsolete now that we have a proper package on conda-forge, and I'll be removing the `conda` folder in a future PR.

conda/xscen/meta.yaml
```diff
@@ -18,34 +18,37 @@ build:
...
     - pip
   run:
-    - python >=3.8
     - cartopy
     - cftime
     - cf_xarray >=0.7.6
-    - clisops >=0.9.5
-    - dask-core
+    - clisops >=0.10
+    - dask
```
Zeitsperre, line 28: I can't remember, what are the differences between `dask` and `dask-core`? Does `dask` vendor `distributed` and other helpers?

environment.yml
```diff
@@ -16,17 +14,18 @@ dependencies:
...
-  - intake
-  - intake-esm >=2022.9.18
+  - h5py
+  - intake-esm >=2023.07.07
   - matplotlib
   - netCDF4
   - numpy
   - pandas >= 2
   - parse
-  - pygeos
```
RondeauG, line 26: On utilise `pygeos` ! https://github.com/Ouranosinc/xscen/blob/94fca2e6190d0e7524d3d1678e46faf0fca51c3c/xscen/aggregate.py#L500

Zeitsperre (changes requested):

conda/xscen/meta.yaml
aulemahal: I think so yeah. Installing `dask` with conda will pull ``distributed`. I did not realize this difference as I only copy-pasted the env list... I can change it back to `-core`. `distributed` is indeed not necessary for xscen.

environment.yml
Zeitsperre: I think the confusion is because pygeos comes bundled with Shapely now (IIRC). But given that we _explicitly_ call it, we should also explicitly mark it in the recipe(s).

aulemahal: Pareil pour:
```python3
import xscen as xs

ds = xr.tutorial.open_dataset("air_temperature")
xs.aggregate.spatial_mean(ds, method='cos-lat')
```

Je suis sur le projet.

aulemahal: @Zeitsperre I think that it's rather that `pygeos` has been _integrated_ into `shapely`. The package still exist, but is not needed anymore, the functions are effectively migrated to the shapely API. In fact, having both `pygeos` and `shapely >= 2` triggers a warning in `geopandas` because of the "ambiguity". 

In theory, replacing `pygeos` by `shapely` works directly. However, this made me realize that the function is broken, so...

PR #259, opened 2023-09-20
Better DATES

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz
- [ ] HISTORY.rst has been updated (with summary of main changes).

Minimal pull request that fixes a bug I just had in the data catalogs.

The "DATES" format flag now needs at least 4 digits to match and a maximum of 15. I had removed that condition earlier, but I believe it was out of laziness to actually think about it. I see no way that a date would have fewer than 4 digits. I had some CORDEX files with a trailing ".1.nc" in the filename, and it shouldn't be recognized as date.

In `date_parser`, also catch `ValueError` when `pd.Timestamp` fails. I didn't know the object would raise that.

No.

### DIFF
--- a/xscen/catutils.py
+++ b/xscen/catutils.py
@@ -90,7 +90,10 @@ def _parse_level(text: str) -> str:
     return text
 
 
-@register_parse_type("datebounds", regex=r"(([\d]+(\-[\d]+)?)|fx)", group_count=3)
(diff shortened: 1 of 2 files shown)


RondeauG (approved):
LGTM!

PR #260, opened 2023-09-25
Better spatial mean

### LINKED ISSUE #94: Improve spatial_mean

aulemahal:
### Pull Request Checklist:
    - This PR fixes #94 

* One can now pass `region='global'` to `spatial_mean`. It will simply be replace by a bbox region covering the whole globe.
* `spatial_mean`, method `xesmf` will automatically segmentize the polygons down to a resolution of 1°, to ensure a correct average.
* Make `out_chunks` available in `regrid_dataset` and `spatial_mean` if using xesmf >= 0.8
* Fix `spatial_mean` with method `xesmf` and a `GeoDataFrame` as shape.

Shouldn't.

### Other information:
Added a test for  `region='global'`.

### DIFF
--- a/xscen/aggregate.py
+++ b/xscen/aggregate.py
@@ -374,7 +374,7 @@ def spatial_mean(
     *,
     spatial_subset: bool = None,
     call_clisops: bool = False,
-    region: dict = None,
+    region: Union[dict, str] = None,
     kwargs: dict = None,
     simplify_tolerance: float = None,
     to_domain: str = None,
@@ -394,10 +394,11 @@ def spatial_mean(
     spatial_subset : bool
         If True, xscen.spatial.subset will be called prior to the other operations. This requires the 'region' argument.
         If None, this will automatically become True if 'region' is provided and the subsetting method is either 'cos-lat' or 'mean'.
-    region : dict
+    region : dict or str
         Description of the region and the subsetting method (required fields listed in the Notes).
         If method=='interp_centroid', this is used to find the region's centroid.
         If method=='xesmf', the bounding box or shapefile is given to SpatialAverager.
+        Can also be "global", for global averages. This is simply a shortcut for `{'name': 'global', 'method': 'bbox', 'lon_bnds' [-180, 180], 'lat_bnds': [-90, 90]}`.
     kwargs : dict
         Arguments to send to either mean(), interp() or SpatialAverager().
         For SpatialAverager, one can give `skipna` here, to be passed to the averager call itself.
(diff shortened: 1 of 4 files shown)


juliettelavoie (commented):
Juste pour bien comprendre `segmentize` c'est que les vertices du polygone sont trop longues et on les sépare en plusieurs et `simplify` c'est que les vertices sont trop courtes, alors on en combine plusieurs ensemble ?

Est-ce que le 1 en argument de `segmentize` c'est aussi en degré?

aulemahal:
Exact! Les deux ont un argument "tolerance" dans les mêmes unités que les coordonnées des polygones. Donc des degrés pour tout nos cas. `simplify` est évidemment un peu plus complexe, y a aussi des tests de validité et autres, mais l'idée c'est en effet l'inverse de `segmentize`.

xscen/aggregate.py
```diff
@@ -632,14 +635,19 @@ def spatial_mean(
...
             raise ValueError("'method' should be one of [bbox, shape].")
 
         kwargs_copy = deepcopy(kwargs)
-        skipna = kwargs_copy.pop("skipna", False)
+        call_kwargs = {"skipna": kwargs_copy.pop("skipna", False)}
+        if "out_chunks" in kwargs:
+            call_kwargs["out_chunks"] = kwargs_copy.pop("out_chunks")
+
+        # Pre-emptive segmentization. Same threshold as xESMF, but there's not strong analysis behind this choice
+        geoms = shapely.segmentize(polygon.geometry, 1)
```
RondeauG, line 650: Am I right in thinking that this will be skipped if we provided a shapefile with a resolution finer than 1°?

RondeauG (approved):
LGTM! I'm happy to see that bug finally resolved 😄

xscen/aggregate.py
aulemahal: It won't do anything to the shape indeed.

`geoms == shapely.segmentize(geoms, 1)` if no geom has any vertex longer than 1.

Issue #169, opened 2023-03-16
Other export formats


RondeauG:
### Addressing a Problem?

For `spirograph`, we might need `save_to_shapefile` or `save_to_geojson`. We might also want `save_to_csv` for some projects.

### Contribution

- [ ] I would be willing/able to open a Pull Request to contribute this feature.

RondeauG:
`save_to_excel` would be useful too and might be better than `save_to_csv`.
@sarahclaude

RondeauG:
@vindelico

PR #262, opened 2023-09-26
Save to excel

aulemahal:
### Pull Request Checklist:
    - This PR fixes some points of #169 

New `io.to_table` and `io.save_to_table` for saving datasets to dataframes / csv / excel / etc. 

This adds support for multi-column and multi-sheet to `ds.to_dataframe()`.

It also supports adding auxiliary coordinates as columns in the output table, beside the data variables. This is actually the most complex part of the code and it might not cover all cases :roll_eyes:...

I also sneaked in a little fix for `save_to_netcdf`, to allow `compute=False`. And I took `coerce_attrs` out of the save function to reduce code duplication.

No.

### Other information:
More testing and doc to come.

### DIFF
--- a/xscen/io.py
+++ b/xscen/io.py
@@ -14,10 +14,12 @@ import xarray as xr
 import zarr
 from rechunker import rechunk as _rechunk
 from xclim.core.calendar import get_calendar
+from xclim.core.options import METADATA_LOCALES
+from xclim.core.options import OPTIONS as XC_OPTIONS
 
 from .config import parse_config
 from .scripting import TimeoutException
-from .utils import season_sort_key, translate_time_chunk
+from .utils import TRANSLATOR, season_sort_key, translate_time_chunk
 
 logger = logging.getLogger(__name__)
 
@@ -26,6 +28,7 @@ __all__ = [
     "clean_incomplete",
     "estimate_chunks",
     "get_engine",
+    "make_toc",
     "rechunk",
     "rechunk_for_saving",
     "save_to_table",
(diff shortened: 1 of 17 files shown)


aulemahal:
The doc issue comes from xclim and has been fixed in https://github.com/Ouranosinc/xclim/pull/1482/commits/c846ca6c02c286f95fe4d57ce796880d8c649d85,

aulemahal:
Does @sarahclaude  or anyone else already have code to add some kind of table of contents to the excel ?

I can write something simple if not.

sarahclaude:
I had done this in previous code;
```
def explication_sheet(path): #creates excel file with path and adds readme sheet
    df1 = pd.DataFrame(data={'column_title': ['experiment_indicator_percentile'],
                             'experiment': ['ssp245 or ssp585'],
                             'percentile': ['10 or 50 or 90'],})

    with pd.ExcelWriter(path, engine='openpyxl') as writer:
        df1.to_excel(writer, sheet_name='readme', index=False)
    return None
```
df1 could be provided by user in the argument. 

And I had this section in the loop that runs through the datasets;
```
df = pd.DataFrame(data={'indicator_abbreviation': list(ds.data_vars),
                                                'indicator_long_name': [ds[v].attrs['long_name'] for v in
                                                                        ds.data_vars]})
df = df.drop_duplicates(subset=['indicator_long_name'])
df['indicator_abbreviation'] = df.apply(lambda row: '_'.join(row[0].split('_')[1:-1]), axis=1)
with pd.ExcelWriter(path, engine='openpyxl', mode='a', if_sheet_exists='overlay') as writer:
    df.to_excel(writer, sheet_name='readme', index=False, startrow=4)
```

xscen/io.py
```diff
@@ -493,6 +490,235 @@ def coerce_attrs(attrs):
...
+        if len(ds) == 1:
+            da = da.isel(variable=0).rename(data=da.variable.values[0])
+
+    def _ensure_list(seq):
+        if isinstance(seq, str):
+            return [seq]
+        return list(seq)
+
+    row = _ensure_list(row or ([d for d in da.dims if d != "variable"]))
+    column = _ensure_list(column or (["variable"] if len(ds) > 1 else []))
```
juliettelavoie, line 596: I think there are issues here with automatically filling the row, colums and sheet. It looks like they are filled even if passed somewhere else.

One example:
I tried with a 2d dataset with dims `time` x `geom` and many variables.
`df= xs.io.to_table(ds1, sheet=['variable'])` fails with 

`ValueError: Repeated dimension names. Got row=['time', 'geom'], column=['variable'] and sheet=['variable'].Each dimension should appear only once.`

aulemahal:
Last commits added a simple toc. It can be localized (column names and long_names when available) with xclim's `metadata_locales` option.

If the `make_toc` function is not enough for the use, one can still pass a DataFrame through the `add_toc` argument. However, I kept it simple by only allowing a single DataFrame, whereas Sarah's example puts two tables in the first page.
I believe the same idea for a more complex TOC could still be implemented like this (in your own script):

```python3
toc_tables = # ... Complex TOC generation...
# Write toc tables to the Content sheet
with pd.ExcelWriter(path, mode='a', if_sheet_exists='overlay') as writer:
    startrow = 0
    for toc_table in toc_tables:
        toc_table.to_excel(writer, sheet_name='Content', startrow=startrow)
        startrow += toc_table.length + 2

xs.io.save_to_table(ds, path, mode='a')  #  mode 'a' to append to file passed to pandas.
```

xscen/io.py
aulemahal: I think my last commit makes better guesses for the dimensions!

juliettelavoie (approved):
I'm not in the new user request group, so I don't know if this fit for purpose. But, I played around with`to_table` with a few of my data and it seems to work well.

sarahclaude (approved):
I think the flexibility of sheet/row/colums is great for future requests and works well!
Could accept strings in coords since all other similar arguments do

PR #266, opened 2023-10-02
Better save_to_zarr

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

* New func ``xs.io.round_bits`` to round a floating point variable up to a number of bits, allowing for a much better compression afterwards.
* New "bitround" arg to `save_to_zarr`  and `save_to_netcdf`, allowing to combine the bit rounding and the saving. The argument can take many forms for more flexibility. 

No.

### Other information:
Default keepbits value is guessed from the variable name, with a default of 12 (0.0122% of rel error). Better guesses should be given in `xs.io.KEEPBITS`, but I have none for now.

### DIFF
--- a/tests/test_io.py
+++ b/tests/test_io.py
@@ -130,7 +130,10 @@ def test_round_bits(datablock_3d):
     dar = xs.io.round_bits(da, 12)
     # Close but NOT equal, meaning something happened
     np.testing.assert_allclose(da, dar, rtol=0.013)
-    assert not (da == dar).any()
+    # There's always a chance of having a randomly chosen number with only zeros in the bit rounded part of the mantissa
+    # Assuming a uniform distribution of binary numbers (which it is not), the chance of this happening should be:
+    # 2^(23 - 12 + 1) / 2^24 = 2^(-12) ~ 0.02 % (but we'll allow 1% of values to be safe)
+    assert (da != dar).sum() > (0.99 * da.size)
 
 
 class TestSaveToZarr:
--- a/tests/test_io.py
+++ b/tests/test_io.py
@@ -116,3 +116,38 @@ class TestToTable:
         np.testing.assert_array_equal(
             tab.loc[("1993", "pr"), ("JFM",)], self.ds.pr.sel(time="1993", season="JFM")
         )
+
+
+def test_round_bits(datablock_3d):
+    da = datablock_3d(
+        np.random.random((30, 30, 50)),
+        variable="tas",
+        x="lon",
+        x_start=-70,
+        y="lat",
+        y_start=45,
+    )
+    dar = xs.io.round_bits(da, 12)
+    # Close but NOT equal, meaning something happened
+    np.testing.assert_allclose(da, dar, rtol=0.013)
+    assert not (da == dar).any()
+
+
+class TestSaveToZarr:
+    @pytest.mark.parametrize(
+        "vname,vtype,bitr,exp",
+        [
+            ("tas", np.float32, 12, 12),
+            ("tas", np.float32, False, None),
+            ("tas", np.int32, 12, None),
+            ("tas", np.int32, {"tas": 2}, "error"),
+            ("tas", object, {"pr": 2}, None),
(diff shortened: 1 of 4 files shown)


xscen/io.py
```diff
@@ -292,12 +297,54 @@ def _coerce_attrs(attrs):
...
+    da.attrs["history"] = history
+    return da
+
+
+def _guess_keepbits(bitround, varname, vartype):
+    # Guess the number of bits to keep depending on how bitround was passed, the var dtype and the var name.
+    if not np.issubdtype(vartype, np.floating) or bitround is False:
+        if isinstance(bitround, dict) and varname in bitround:
+            raise ValueError(
+                f"A keepbits value was given for variable {varname} even though it is not of a floatig dtype."
```
coxipi, line 329: typo 
```suggestion
            raise ValueError(
                f"A keepbits value was given for variable {varname} even though it is not of a floating dtype."
```

xscen/io.py
```diff
@@ -314,6 +361,11 @@ def save_to_netcdf(
         Spatial dimensions can be generalized as 'X' and 'Y', which will be mapped to the actual grid type's
         dimension names.
         Rechunking is only done on *data* variables sharing dimensions with this argument.
+    bitround : bool or int or dict
+      If not False, float variables are bit-rounded by dropping a certain number of bits from their mantissa, allowing for a much better compression.
+      If an int, this is the number of bits to keep for all float variables.
+      If a dict, a mapping from variable name to the number of bits to keep.
+      If True, the number of bits to keep is guessed based on the variable's name, defaulting to 12, which yields a relative error of 0.012%.
```
coxipi, line 368: To be more precise, we should say something like: 
> The **maximum possible absolute relative error** will be "0.0122 ... %", 

as this is an upper bound.  For simplicity, we can drop the "absolute" which can be implicit, but otherwise I would write something like:  
```suggestion
      If True, the number of bits to keep is guessed based on the variable's name, defaulting to 12, which yields a relative error below 0.0125%.
```
or `below 0.013%` if we prefer keeping two significant digits.

tests/test_io.py
```diff
@@ -116,3 +116,38 @@ def test_normal(self):
...
+        variable="tas",
+        x="lon",
+        x_start=-70,
+        y="lat",
+        y_start=45,
+    )
+    dar = xs.io.round_bits(da, 12)
+    # Close but NOT equal, meaning something happened
+    np.testing.assert_allclose(da, dar, rtol=0.013)
+    assert not (da == dar).any()
```
coxipi, line 133: This last check could randomly fail if the `np.random.random` generates numbers with a lots of zeroes in the trailing bits.  Say you have a number in your array whose mantissa bits are
1110101**0000000000000000**
i.e. only the 7 first mantissa bits form a non-trivial chain, then keeping 12 bits will result in the same number. 

This is probably very unlikely to get numbers like this with `np.random.random`

xscen/io.py
```diff
@@ -292,12 +297,54 @@ def _coerce_attrs(attrs):
...
+                f"A keepbits value was given for variable {varname} even though it is not of a floatig dtype."
+            )
+        return None
+    if bitround is True:
+        return KEEPBITS[varname]
+    if isinstance(bitround, int):
+        return bitround
+    if isinstance(bitround, dict):
+        return bitround.get(varname, KEEPBITS[varname])
+    return None
```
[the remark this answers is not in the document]
aulemahal: Indeed the defaultdict is the real guessing stuff, it just doesn't have any guesses in yet... My idea is that we would add some variable names with non-default keep bits values if we come to the conclusion that these variables merit something else than 12.

Good point about the naming, it is a bit confusing indeed.

tests/test_io.py
aulemahal: If my computations are correct, the probability of obtaining a number with the 12 last bits of mantissa all zeroes is 2^-12, which is small, but possible! I relaxed the test.

coxipi (approved):
Nice, looks good to me!

PR #282, opened 2023-11-01
Multiple small fixes to docstrings and typing

RondeauG:
### Pull Request Checklist:
    - No.

* Removed a few outdated imports.
* Typing was added to multiple functions when it was missing.
* Output types were added to multiple functions when it was missing.
* Docstrings was added to multiple functions when it was missing.
* Both `str` and `os.PathLike` should be explicitly supported where relevant.
* Minor bugfixes when I saw them.
* Module descriptions were added when missing.
* To make things readable:
  * As much as possible, the type of data in lists was defined.
  * The type of data in dictionaries was not.

- It should not.

### Other information:

- Pycharm recommended that I add `ABC` to `DataCatalog/ProjectCatalog`. It this right? @aulemahal 
- The `init` and documentation of `DataCatalog/ProjectCatalog` should be reworked to be easier to understand, but this is out-of-scope for this PR. For example we use `*args`, but AFAIK there is only one possible input argument, `obj`.


xscen/regrid.py
```diff
@@ -1,10 +1,9 @@
-# noqa: D100
+"""Regrid datasets using xESMF."""
```
juliettelavoie, line 1: maybe in the fututre it wont always be xesmf ?

juliettelavoie (approved):

xscen/catalog.py
```diff
@@ -240,10 +256,13 @@ def _find_unique(series):
...
             return data.apply(_find_unique, result_type="reduce").to_dict()
 
-    def unique(self, columns: Union[str, list] = None):
+    def unique(self, columns: Union[str, Sequence[str]] = None):
         """Return a series of unique values in the catalog.
 
-        Subsets on a columns list if specified.
+        Parameters
+        ----------
+        columns : str or list of str, optional
```
Zeitsperre, line 264:
```suggestion
        columns : str or sequence of str, optional
```

xscen/catalog.py
```diff
@@ -283,7 +302,14 @@ def search(self, **columns):
             )
         return cat
 
-    def drop_duplicates(self, columns: Optional[list[str]] = None):  # noqa: D102
+    def drop_duplicates(self, columns: list[str] = None):
```
Zeitsperre, line 305:
```suggestion
    def drop_duplicates(self, columns: Optional[list[str]] = None):
```

xscen/catalog.py
```diff
@@ -364,8 +397,8 @@ def exists_in_cat(self, **columns):
 
     def to_dataset(
         self,
-        concat_on: Optional[Union[list[str], str]] = None,
-        create_ensemble_on: Optional[Union[list[str], str]] = None,
+        concat_on: Union[list[str], str] = None,
+        create_ensemble_on: Union[list[str], str] = None,
```
Zeitsperre, line 401:
```suggestion
        create_ensemble_on: Optional[Union[list[str], str]] = None,
```

xscen/catalog.py
```diff
@@ -496,18 +529,18 @@ def create(
...
 
         Creates the json from default :py:data:`esm_col_data` and an empty csv file.
 
         Parameters
         ----------
-        filename : PathLike
+        filename : os.PathLike or str
           A path to the json file (with or without suffix).
-        project : dict-like
+        project : dict, optional
```
Zeitsperre, line 543: I'm confused, can `project` be `None` or not?

xscen/catalog.py
```diff
@@ -585,15 +618,15 @@ def __init__(
...
-        df : str, dict
+        df : str or dict
             If str, this must be a path or URL to a catalog JSON file.
             If dict, this must be a dict representation of an ESM catalog.  See the notes below.
         create : bool
             If True, and if 'df' is a string, this will create an empty ProjectCatalog if none already exists.
-        project : dict-like
-            Metadata to create the catalog, if required.
-        overwrite : bool
+        overwrite : bool, optional
```
Zeitsperre, line 626: Why not set this to `False` by default? This means it accepts `True | False | None`.

xscen/catalog.py
```diff
@@ -615,14 +648,12 @@ def __init__(
...
-                pd.DataFrame,
-                pd.Series,
-                Sequence[pd.Series],
-            ]
+        df: Union[
+            DataCatalog,
+            intake_esm.esm_datastore,
+            pd.DataFrame,
+            pd.Series,
+            Sequence[pd.Series],
```
Zeitsperre, line 656:
```suggestion
        df: Optional[Union[
            DataCatalog,
            intake_esm.esm_datastore,
            pd.DataFrame,
            pd.Series,
            Sequence[pd.Series]],
```

xscen/catalog.py
```diff
@@ -822,17 +853,22 @@ def _build_id(element: pd.Series, columns: list[str]):
 
 
 def generate_id(
-    df: Union[pd.DataFrame, xr.Dataset], id_columns: Optional[list] = None
-):  # noqa: D401
-    """Utility to create an ID from column entries.
+    df: Union[pd.DataFrame, xr.Dataset], id_columns: list = None
```
Zeitsperre, line 856: ibid.

xscen/catutils.py
```diff
@@ -104,13 +106,26 @@ def _parse_datebounds(text: str) -> tuple[str, str]:
 
 
 def _find_assets(
-    root: Union[str, Path],
+    root: Union[str, os.PathLike],
     exts: set[str],
     lengths: set[int],
-    dirglob: Optional[str] = None,
+    dirglob: str = None,
```
Zeitsperre, line 112: This change means that passing `dirglob = None` explicitly is no longer allowed.

Zeitsperre (changes requested):
Good changes overall. I think the `Optional` usage needs to be clarified here. Will be good to go afterwards!

EDIT: To reiterate, the `Optional` serves both the IDE/user and helps with the static type checking that would come from [mypy](https://www.mypy-lang.org/). We don't currently have any projects that are 100% `mypy`-compliant, but the more we implement now, the easier it will be in the future.

Why use `mypy`? https://python.plainenglish.io/mypy-your-secret-weapon-for-code-brilliance-a376d15c463c

xscen/aggregate.py
```diff
@@ -47,26 +47,26 @@ def climatological_mean(
     window: int = None,
     min_periods: int = None,
     interval: int = 1,
-    periods: list = None,
-    to_level: str = "climatology",
+    periods: Union[list[str], list[list[str]]] = None,
+    to_level: Optional[str] = "climatology",
```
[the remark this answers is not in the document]
Zeitsperre: Having re-read the PR description, I still think that we don't lose much by keeping the `Optional`s around (these will be replaced with `xyz | None` in Python3.10 (or if we adopt `from __future__ import annotations`.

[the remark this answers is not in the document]
RondeauG: I don't mind adding back those if you say that `Optional` should be everywhere. From my testing, PyCharm had no problem understanding that you can explicitly use `None`, when this was the default.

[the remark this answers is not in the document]
Zeitsperre: PyCharm type checking isn't as exhaustive/extensive as `mypy`. Once we indicate to `pyupgrade` that we want to move to Python3.10 typing, this will look much nicer without any intervention needed on our end.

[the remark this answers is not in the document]
RondeauG: Alright, in that case I'll go with `Optional` everywhere relevant.

xscen/catalog.py
```diff
@@ -135,22 +136,37 @@ def _parse_dates(elem):
 """Kwargs to pass to `pd.read_csv` when opening an official Ouranos catalog."""
 
 
-class DataCatalog(intake_esm.esm_datastore):
-    """
+class DataCatalog(intake_esm.esm_datastore, ABC):
```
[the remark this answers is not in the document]
RondeauG: Honestly not sure, this is a suggestion from PyCharm (which is why I tagged Pascal). Without it, I get a suggestion saying:

```
Class DataCatalog must implement all abstract methods.

Solution: Add 'abc.ABC' to superclasses
```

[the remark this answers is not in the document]
aulemahal: AJA ce qu'était une abstract method. Ça l'air que c'est un moyen pour une classe de dire « je ne suis pas complète, mes enfants doivent implémenter les méthodes .foo et .bar », où .foo et .bar sont des abstract methods. Ainsi donc, la classe en question n'a pas le droit d'être instantiée (pas de `obj = LaClasse()`). Il faut en faire une sous-classe et implémenter les méthodes .foo et .bar avant.

Mais je pense que PyCharm est tombé sur la tête, parce que je ne vois pas aucune de ces abstract methods dans `DataCatalog` et ces ancêtres, Je pense qu'on peut l'enlever.

xscen/catalog.py
```diff
@@ -483,7 +516,7 @@ def preprocess(ds):
         return ds
 
 
-class ProjectCatalog(DataCatalog):
+class ProjectCatalog(DataCatalog, ABC):
```
aulemahal, line 519: Pareil, surtout que si ProjectCatalog hérite de DataCatalog, il aurait hérité de facto de ABC!

xscen/catalog.py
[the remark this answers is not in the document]
RondeauG: Alright, dans ce cas je vais enlever.

xscen/catalog.py
RondeauG: This had been changed to be in line with my statement on Optional, but I've put it back everywhere.

xscen/catalog.py
RondeauG: Indeed, this made no sense. I removed the option to put `None` here (especially since it wasn't actually supported by `ProjectCatalog.create()`.

xscen/aggregate.py
```diff
@@ -372,13 +372,13 @@ def spatial_mean(
...
     *,
-    spatial_subset: bool = None,
-    call_clisops: bool = False,
-    region: Union[dict, str] = None,
-    kwargs: dict = None,
-    simplify_tolerance: float = None,
-    to_domain: str = None,
-    to_level: str = None,
+    spatial_subset: Optional[bool] = None,
+    call_clisops: Optional[bool] = False,
```
Zeitsperre, line 376: Just to be certain, `call_clisops = None` is valid?

xscen/catutils.py
```diff
@@ -162,20 +177,20 @@ def _compile_pattern(pattern: str) -> parse.Parser:
...
-    path: os.PathLike,
-    root: os.PathLike,
+    path: Union[os.PathLike, str],
+    root: Union[os.PathLike, str],
     patterns: list[Union[str, parse.Parser]],
     read_from_file: Optional[Union[list[str], dict]] = None,
     attrs_map: Optional[dict] = None,
     xr_open_kwargs: Optional[dict] = None,
-):
+) -> Union[dict, None]:
```
Zeitsperre, line 186: Is `Optional[dict]` a valid return signature?

xscen/utils.py
```diff
@@ -522,11 +533,29 @@ def get_cat_attrs(
 @parse_config
 def maybe_unstack(
     ds: xr.Dataset,
-    coords: str = None,
-    rechunk: bool = None,
+    coords: Optional[str] = None,
+    rechunk: Optional[bool] = None,
```
Zeitsperre, line 537: Same here

Zeitsperre (approved):
Lots of work here. looking good!

xscen/aggregate.py
RondeauG: Yes, this is intended:
```
If True, xscen.spatial.subset will be called prior to the other operations. This requires the 'region' argument.
If None, this will automatically become True if 'region' is provided and the subsetting method is either 'cos-lat' or 'mean'.
```

xscen/catutils.py
RondeauG: I think so. I'll change it.

xscen/utils.py
RondeauG: Good catch. After speaking with Juliette, this turns out to be a type error. It should be `Optional[dict]`, not `bool`.

PR #284, opened 2023-11-07
Updated show_versions() list and warning to devs

### LINKED ISSUE #281: More robust `show_versions`
### Generic Issue

* xscen version: 0.7.17

### Description
`xscen.utils.show_versions` has an hard-coded list of dependencies, which isn't great when they change. Ideally, the code should look for the real dependencies.

### Possible solution
Taken from elsewhere, but could be adapted to work here.
```
def check_pip_version(pkgs):
    '''returns a dictionary of packages {pkg_name:version}, as returned by pip list.'''
    try:
        from importlib import metadata
    except ImportError: # for Python<3.8, or bokeh < 3.0.2
        import importlib_metadata as metadata
    return {k:metadata.version(k) for k in pkgs}
```

RondeauG:
### Pull Request Checklist:
    - This PR fixes #281

* Reordered the list of packages in `xs.utils.show_versions()` so it matches `environment.yml` more closely, for easier comparison between the two. I also added a few new packages that weren't in the list, such as `babel`.
* Updated the warning in our environment files to be more exhaustive on the list of files to update.

- No.


xscen/utils.py
```diff
@@ -1183,35 +1183,45 @@ def show_versions(
     """
     if deps is None:
         deps = [
```
Zeitsperre, line 1185: OK, so I know it's a bit silly, but I would reverse this list here since the way that the function is written, it reports from last to first. Placing `xscen` last would be good as well.

Zeitsperre: Could even just use a `deps.reverse()` for readability

Zeitsperre (approved):

xscen/utils.py
RondeauG: There's actually a `sorted` in the `xclim` function, so the order doesn't matter:
```
INSTALLED VERSIONS
------------------
python: 3.10.12
babel: 2.12.1
cartopy: 0.22.0
cf_xarray: 0.8.4
cftime: 1.6.2
clisops: 0.11.0
dask: 2023.8.1
fastprogress: 1.0.3
flox: 0.7.2
fsspec: 2023.6.0
geopandas: 0.13.2
h5netcdf: 1.2.0
h5py: 3.9.0
intake: 0.7.0
intake_esm: 2023.7.7
matplotlib: 3.7.2
nc-time-axis: None
netCDF4: 1.6.4
numpy: 1.24.4
pandas: 2.1.0
parse: 1.19.1
pyarrow: 13.0.0
pydantic: 2.3.0
pyyaml: None
rechunker: 0.5.2
requests: 2.31.0
shapely: 2.0.1
sparse: 0.14.0
toolz: 0.12.0
xarray: 2023.8.0
xclim: 0.45.0
xcollection: None
xesmf: 0.7.1
xscen: 0.7.19-beta
yaml: 6.0.1
zarr: 2.16.1
Anaconda-based environment: yes

```

Zeitsperre: Looks like I'm the one that added that. Well alright then XD

Issue #298, opened 2023-12-06
Input requirements break `xs.ensemble_stats`


RondeauG:
### Setup Information

- xscen version: 0.7.21
- xclim version: 0.47

### Description

Currently, `xs.ensemble_stats` tries to handle multiple functions in `xclim.ensembles` that do not necessarily have compatible input requirements. Some examples:

- `mean_std_max_min` and `ensemble_percentiles` compute their statistics along `realization`. Thus, `time` should already have been aggregated (e.g. through a 30-year mean). --> Works well with horizons.
- `robustness_factors` with `ref=None` also only works across `realization`, and considers the input to be pre-computed deltas where `time` has been aggregated. --> Works well with horizons.
- `robustness_factors` with `ref=xr.DataArray` **needs to have a `time` axis and will average `time` itself**. I think this is because some tests need daily or yearly data. --> Completely breaks the concept of horizons.
- Other functions in `xclim.ensembles` can also be called in theory, but diverge so much from what we currently have implemented that they probably don't work at all.

### Additional context

`xs.ensemble_stats` performs no checks and never warns the user if the list of tests is incompatible with the data provided. 

At minimum, we should add that. Better yet, we should try to find a way to support that, or split `xs.ensemble_stats` in multiple functions.

### Contribution

- [X] I would be willing/able to open a Pull Request to address this bug.

PR #299, opened 2023-12-06
Support robustness functions in `xs.ensemble_stats`

RondeauG:
### Pull Request Checklist:
    - No.

* You can now directly provide an ensemble dataset to `xs.ensemble_stats`.
* `xs.ensemble_stats` now supports the robustness-related functions from `xclim`.
* Small bugfix in `publish_release_notes`

* Removed explicit support for `ref` as it was implemented, since it actually was completely broken... See #298.

### DIFF
--- a/xscen/ensembles.py
+++ b/xscen/ensembles.py
@@ -150,6 +150,10 @@ def ensemble_stats(  # noqa: C901
             # FIXME: This can be removed once change_significance is removed.
             #  It's here because the 'ref' default was removed for change_significance in xclim 0.47.
             stats_kwargs.setdefault("ref", None)
+            if (stats_kwargs.get("ref") is not None) and len(statistics.keys()) > 1:
+                raise ValueError(
+                    f"The input requirements for '{stat}' when 'ref' is specified are not compatible with other statistics."
+                )
 
             # These statistics only work on DataArrays
             for v in ens.data_vars:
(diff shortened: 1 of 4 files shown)


xscen/ensembles.py
```diff
@@ -59,62 +66,143 @@ def ensemble_stats(
...
+            "change_significance",
+            "robustness_fractions",
+            "robustness_categories",
+        ]:
+            # FIXME: This can be removed once change_significance is removed.
+            #  It's here because the 'ref' default was removed for change_significance in xclim 0.47.
+            stats_kwargs.setdefault("ref", None)
+            if (stats_kwargs.get("ref") is not None) and len(statistics.keys()) > 1:
+                raise ValueError(
+                    f"The input requirements for '{stat}' when 'ref' is specified are not compatible with other statistics."
```
juliettelavoie, line 155: so it's okay to put ref, but only if we only request one stat ?

What if we want `robustness_categories`, then we have to pass `'robustness_fractions'`, which means we can't use ref. Is that right ?

maybe a bit more guiding on ref would help in the docstring.

RondeauG: My error, it should have been `len(statistics_to_compute)`. Basically if you want to use `ref`, then you can't call other statistics such as `ensemble_percentiles`.

I've also added more details on `ref`.

juliettelavoie (approved):
LGTM!

PR #304, opened 2023-12-14
Do not fail because of Linkcheck, add ignored links

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #xyz

* Added some links to ignore
* Configured ReadTheDocs to not fail build if Linkcheck step fails
* Moved notebooks into an "Examples" section

No.


juliettelavoie (approved):
ah! thanks !

Zeitsperre:
@aulemahal 

It looks like your latest notebook is failing. Shall I leave the fix up to you here?

aulemahal:
Woups, this notebook is not supposed to run. I'll deactivate it.

I kept it in the repo to show how the results were obtained, and to make it easy to manually update the timeseries, periodically. But it makes reference to data inside ouranos and this link for NOAA's dataset changes each month.

aulemahal:
@Zeitsperre I add the "NO_COLOR" env var to the admin panel of RTD, should make the online output easier to read!

Otherwise, the only warning I can find in the last build is coming from Clisops... Or I missed something else ?

Zeitsperre:
@aulemahal

The final warning in the docs build is coming from the fact that the notebook is not explicitly ignored. (`UserWarnings` and other Python-specific warnings don't count). I'm on it.

aulemahal:
Oh I see! Thanks!

Zeitsperre:
Errors are finally all addressed. I've restructured the notebooks for flow; If this isn't what we want, I can revert the changes.

docs/notebooks/index.rst
```diff
@@ -0,0 +1,13 @@
...
+.. toctree::
+    :maxdepth: 1
+
+    1_catalog
+    2_getting_started
+    3_diagnostics
+    4_ensemble_reduction
+    5_warminglevels
+    6_config
+    7_global_tas_average_obs
```
juliettelavoie, line 13: I don't think we want that one in the doc? It wasn't there before.

docs/notebooks/index.rst
```diff
@@ -0,0 +1,13 @@
+Examples
```
juliettelavoie, line 1: I don't understand why the need for the new files ?

Zeitsperre: Aesthetics. It gives the notebooks a nested menu. If you aren't a fan, I'll remove it.

docs/notebooks/index.rst
Zeitsperre: If we don't want it, I'll remove it.

docs/notebooks/index.rst
juliettelavoie: ah ok! I'm good with the nested menu.

docs/notebooks/index.rst
juliettelavoie: I would remove, it doesn't use xscen. It is used to build xscen. @aulemahal did you want it in the doc?

aulemahal: I put it there so it's not lost and we can reproduce the work, but it is not meant to be in the main documentation!

docs/notebooks/index.rst
```diff
@@ -0,0 +1,12 @@
...
+========
+
+.. toctree::
+    :maxdepth: 1
```
aulemahal, line 5:
```suggestion
    :maxdepth: 1
    :numbered:
```
Suggestion, so it makes the documentation structure even clearer, with numbered sections.

aulemahal (approved):
I approve this.

PR #319, opened 2024-01-29
Update cookiecutter to new standard via cruft

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #xyz

* Updates the cookiecutter with the newest conventions and workflows

Boilerplate documentation is largely unchanged. Workflows are now more a bit more elegant, including automatic labelling, warnings about unsafe changes to workflows, security-related changes, etc.

`actions-versions-updater.yml` has been replaced with `Dependabot` (it's just better).

There's support for evaluating the OpenSSF Scorecard (this can be disabled if we want).

Code formatting tools are now hard-pinned. These need to be kept in sync with changes from `pre-commit`. `Dependabot` should do this task automatically via Pull Requests.

Versioning scheme is now SemVer 2.0-compliant:
* If the version doesn't end in `-dev` or `-dev.##`, `$ bump-my-version bump patch` will be called. This will set the version at `X.Y.Z+1-dev`. Otherwise, `$ bump-my-version bump build` will be called. This is all automated by the `bump-version.yml`.

When the version is ready for a release, it's up to the maintainer to call the following:
* `$ bump-my-version bump release` (for a patch release; i.e. `1.2.0` → `1.2.1`) or
* `$ bump-my-version bump minor` then `$ bump-my-version bump release` (for a minor release; i.e. `1.2.0` → `1.3.0`)

There is now also support for Documentation locales (translations) via `.po` files. These can be staged with `$ make initialize-translations`, whereby `.po` files will be generated and French (or other) language strings can be added.
    * **I'm not sure how this will interact with existing translation code. Might be worth it to leave this deactivated**

### Other information:

https://github.com/Ouranosinc/cookiecutter-pypackage/pull/30


.github/workflows/main.yml
```diff
@@ -5,11 +5,13 @@ on:
     branches:
       - main
     paths-ignore:
+      - .cruft.json
       - CHANGES.rst
       - README.rst
       - pyproject.toml
       - setup.cfg
       - setup.py
+      - tests/test_xscen.py
```
aulemahal, line 14: Why ?

tox.ini
```diff
@@ -2,11 +2,11 @@
 min_version = 4.0
 envlist =
     lint
-    py{39,310,311}-esmpy
+    py{39,310,31,312}-esmpy
```
aulemahal, line 5:
```suggestion
    py{39,310,311,312}-esmpy
```
I don't think we support Python 3.1 ;)

aulemahal (approved):
Looks good!

.github/workflows/main.yml
Zeitsperre: There is one test that changes when we run `bump-my-version`. We shouldn't need to restart our tests when we bump the version, but maybe that's a good thing to do? I'm open to removing this.

tox.ini
Zeitsperre: Oof. Now that's a fatfinger typo.

.github/workflows/main.yml
Zeitsperre: @aulemahal 

image

aulemahal: Hmm I see!

I suggest removing either:
- the version number from the tested line
- the test file itself.

I don't see the point of that "smoke test" now that we have a minimal test suite. 
 @RondeauG  ?

RondeauG: I agree that we could remove that test if it complicates things!

Zeitsperre: Smoke tests removed!

PR #337, opened 2024-02-06
Make xESMF optional

aulemahal:
### Pull Request Checklist:
    - This PR fixes hydrologie/xhydro#81
- [x] (If applicable) Tests have been updated.

* Make xesmf optional:
    + Remove it from the dependencies in "pyproject.toml". It is now an "extra" dependency. I did not remove it from the environment.yml files as the issue doesn't occur with conda. 
    + Add a few "try: import" throughout the package and raise on call time when xesmf is not available.
    + Skip the tests that require xesmf, if not present.
    + Edit the tox config to remove any usage of conda, since it is now not needed.

No.

### Other information:
@Zeitsperre I may have been heavy handed in editing the workflow and tox configuration. My idea was that the "test-pypi" didn't need xesmf/esmf anymore and so I removed any mention of it and of mamba/conda. So it can really be a "test-pypi" ci.


aulemahal:
Yes ça marche!

Issue with Python 3.9 is because of the latest conda package of intake-esm. They dropped 3.9 but the conda recipe didn't know about that.

RondeauG (approved):
Thanks! I think this will make things much easier.

aulemahal:
CI fails because the misconfigured conda package still exists in mamba's cache and in the conda-forge repo. But it did pass when I pinned intake-esm.

Coverage for Pypi runs are decreased, which is normal since we don't test any function using xesmf. But the conda converage is stable. @Zeitsperre , what should we do ? Accept the decrease ? Stop reporting coverage for those tests ?

But otherwise, this PR is good to go.

tox.ini
Zeitsperre, line 1: Any way you could pull back a few of these changes? Having the option to install `esmpy` with `tox` is nice to have. The `conda_*` fields are primarily there in the event that `tox-conda` comes to support `tox>=4.0` (they do nothing at the moment).

The only change that really is needed is the removal of `-esmpy` from the `envlist` (the default environments to check when calling `tox` without `-e`).

.github/workflows/main.yml
Zeitsperre, line 1: This looks fine to me.

Zeitsperre:
> @Zeitsperre , what should we do ? Accept the decrease ? Stop reporting coverage for those tests ?

I think accepting the one-time decrease is fine. So long as the overall coverage doesn't go down (`test-pypi` + `test-conda`), that's what matters.

.github/workflows/main.yml
```diff
@@ -142,10 +147,6 @@ jobs:
           create-args: >-
             mamba
             python=${{ matrix.python-version }}
-      - name: Downgrade intake-esm
-        if: matrix.python-version == '3.9'
-        run: |
-          micromamba install -y -c conda-forge intake-esm=2023.11.10
```
Zeitsperre, line 153: This is needed for the new `intake-esm`

aulemahal: woups sorry I didn't check the git revert carefully!

PR #340, opened 2024-02-12
Prepare 0.8.2 release

RondeauG:

* Since this makes it easier to install through `pip`, it's worth releasing on PyPI.


RondeauG:
An update ~2 weeks ago releave the `-beta`, so no need to bump versions for this one.

aulemahal (approved):

PR #366, opened 2024-03-11
Add biasadjust tests

juliettelavoie:
### Pull Request Checklist:
    - This PR fixes #xyz
- [ ] CHANGES.rst has been updated (with summary of main changes).

* add bias adjust tests
* fixed a few bugs:
  * Removed train from the inside the period for loop. Only need to do it once and bugged with group if inside loop.
  * fix `_add_preprocessing_attr` that was not actually modifying the attrs
  * Change way of defining default group

*`adjust` won't fail if dict group and many periods.
*`preprocessing_attrs` will actually be added
* If pass `group=False`, it will actually be False.

### Other information:

1. Is it ok to remove the TODOs ? I'm not sure exactly what they were referring to, I don't think they are relevant anymore.
2. I suggest removing `moving_yearly_window` I am not actually using the args for info-crue cmip6 and the function has changed name in xclim.

### DIFF
--- a/tests/test_biasadjust.py
+++ b/tests/test_biasadjust.py
@@ -1,0 +1,354 @@
+import dask
+import numpy as np
+import pytest
+import xarray as xr
+import xclim as xc
+from conftest import notebooks
+from xclim.testing.helpers import test_timeseries as timeseries
+
+import xscen as xs
+
+xc.set_options(
+    sdba_encode_cf=False
+)  # FIXME: A temporary bug fix waiting for xclim 0.49
+
+try:
+    import xesmf as xe
+except ImportError:
+    xe = None
+
+
+class TestTrain:
+    dref = timeseries(
+        np.ones(365 * 3), variable="tas", start="2001-01-01", freq="D", as_dataset=True
+    )
+
+    dhist = timeseries(
+        np.concatenate([np.ones(365 * 2) * 2, np.ones(365) * 3]),
+        variable="tas",
+        start="2001-01-01",
+        freq="D",
+        as_dataset=True,
+    )
+    dhist.attrs["cat:xrfreq"] = "D"
+    dhist.attrs["cat:domain"] = "one_point"
+    dhist.attrs["cat:id"] = "fake_id"
+
+    @pytest.mark.parametrize(
+        "var, period",
+        [("tas", ["2001", "2002"]), (["tas"], ["2003", "2003"])],
+    )
+    def test_basic_train(self, var, period):
+        out = xs.train(self.dref, self.dhist, var=var, period=period)
+
+        assert out.attrs["cat:xrfreq"] == "D"
+        assert out.attrs["cat:domain"] == "one_point"
+        assert out.attrs["cat:id"] == "fake_id"
+        assert out.attrs["cat:processing_level"] == "training_tas"
+
+        assert "dayofyear" in out
+        assert "quantiles" in out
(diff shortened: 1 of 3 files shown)


tests/test_biasadjust.py
```diff
@@ -0,0 +1,345 @@
...
+import xscen as xs
+
+xc.set_options(
+    sdba_encode_cf=False
+)  # FIXME: A temporary bug fix waiting for xclim 0.49
+
+try:
+    import xesmf as xe
+except ImportError:
+    xe = None
```
RondeauG, line 17: I don't think that we use xESMF anywhere in this file?

tests/test_biasadjust.py
```diff
@@ -0,0 +1,345 @@
...
+    #     )
+    #
+    #
+    #     out = xs.adjust(self.dtrain,
+    #                     self.dsim,
+    #                     periods=['2001', '2006'],
+    #                     xclim_adjust_args={'scen': scen,
+    #                                        'frac': 0.25},
+    #                     moving_yearly_window={'window': 2, 'step': 1},
+    #                     )
```
RondeauG, line 345: Still useful?

xscen/biasadjust.py
```diff
@@ -15,6 +15,7 @@
 from .utils import minimum_calendar, standardize_periods
 
 logger = logging.getLogger(__name__)
+xc.set_options(sdba_encode_cf=True, sdba_extra_output=False)
```
RondeauG, line 18:
```suggestion
xc.set_options(sdba_encode_cf=True, sdba_extra_output=False)  # FIXME: Temporary fix. Remove once xclim v0.49 is available.
```

RondeauG (approved):

xscen/biasadjust.py
juliettelavoie: hum pas mal sur qu'on veut garder ça. Le temporary  fix c'est quand on le met False, il me semble. @aulemahal ?

tests/test_biasadjust.py
RondeauG: I haven't used that functionality, and reading the Docstrings, I'm not sure what it actually does... Is it there to optimize calculations for big datasets?

In any case, no strong opinion.

xscen/biasadjust.py
RondeauG: My bad, tu as raison.

tests/test_biasadjust.py
juliettelavoie: It's there if you need to do the adjustement on a rolling X-year window (instead of on the whole simulation period like Detrended Quantile mapping). For ExtremeValues in info-crue-cmip6, I construct the window outside of xscen. I think I probably meant to modify it to use xscen eventually, but I never did.

Now, the function has a new name in xclim (`stack_periods`) and so xscen should probably be updated if we want to keep it. But, it now feels like a very specific case  and I am wondering if it needs to be in xscen.

RondeauG: Gotcha! Still no strong opinion on it, though, especially if it has never actually been useful.

Although if we decide to keep it, I'd fix it to call the new xclim function.

xscen/biasadjust.py
aulemahal: Exact. J'avoue que je n'ai pas retesté l'impact sur la performance depuis un bout, mais à l'époque ça aidait pas mal le scheduler de pré-encoder les cftime.

PR #380, opened 2024-04-11
Address a few FutureWarnings

RondeauG:
### Pull Request Checklist:
    - This PR fixes #xyz

* Addresses a few FutureWarning I encountered recently:
  * `groupby` will change the default to `observed=True`. I think that our implementation here does not care about `observed`, even if we use categoricals, but I'm not 100% sure. We could use `observed=False` to ensure no breaking change.
    * https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.groupby.html
    * https://towardsdatascience.com/be-careful-when-using-pandas-groupby-with-categorical-data-type-a1d31f66b162
  * Changed a few of the old `pandas` codes that were missed.
  * Changed `pd.unique` to `np.unique`
    * `pd.unique with argument that is not not a Series, Index, ExtensionArray, or np.ndarray is deprecated and will raise in a future version.`
    * `intake_esm` no longer spams the warning about `applymap`, so our fix was removed. It still has the "observed=True" spam, however.
    * Changed an implementation of inplace modifications to a DataFrame.
      * `FutureWarning: A value is trying to be set on a copy of a DataFrame or Series through chained assignment using an inplace method. The behavior will change in pandas 3.0. This inplace method will never work because the intermediate object on which we are setting values always behaves as a copy. For example, when doing 'df[col].method(value, inplace=True)', try using 'df.method({col: value}, inplace=True)' or df[col] = df[col].method(value) instead, to perform the operation inplace on the original object.`
  * Added a temporary fix for the `flox` spam in the documentation.

- To avoid breaking changes, 'Y' and 'M' are still allowed in `date_parser`, so no.

### Other information:

-

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -21,6 +21,7 @@ Internal changes
 * Added more tests. (:pull:`366`, :pull:`367`, :pull:`372`).
 * Refactored ``xs.spatial.subset`` into smaller functions. (:pull:`367`).
 * An `encoding` argument was added to ``xs.config.load_config``. (:pull:`370`).
+* Various small fixes to the code to address FutureWarnings. (:pull:`380`).
 
 Bug fixes
 ^^^^^^^^^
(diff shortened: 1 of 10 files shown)


aulemahal (approved):
Looks good to me.

PR #382, opened 2024-04-12
skip missing check if no time in variable

RondeauG:
### Pull Request Checklist:
    - This PR fixes #xyz

* Added a check to make sure that the `time` dimension exists for the variable before checking for missing data.

* No.

### DIFF
--- a/xscen/diagnostics.py
+++ b/xscen/diagnostics.py
@@ -262,11 +262,17 @@ def health_checks(  # noqa: C901
             for method, kwargs in missing.items():
                 kwargs.setdefault("freq", "YS")
                 for v in ds.data_vars:
-                    ms = getattr(xc.core.missing, method)(ds[v], **kwargs)
-                    if ms.any():
-                        _error(
-                            f"The variable '{v}' has missing values according to the '{method}' method.",
-                            "missing",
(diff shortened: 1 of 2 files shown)


xscen/diagnostics.py
```diff
@@ -262,11 +262,17 @@ def _message():
...
+                        ms = getattr(xc.core.missing, method)(ds[v], **kwargs)
+                        if ms.any():
+                            _error(
+                                f"The variable '{v}' has missing values according to the '{method}' method.",
+                                "missing",
+                            )
+                    else:
+                        warnings.warn(
+                            f"Variable '{v}' has no time dimension. The missing data check will be skipped.",
+                            UserWarning,
```
aulemahal, line 275: Feels like this could rather be an INFO logging call ?
Or you think it is abnormal that the user requested a missing check when there is no time dimension ?

aulemahal (approved):
Looks good.

xscen/diagnostics.py
RondeauG: OK to change that to a logging call. The issue here is that it loops through all variables in the dataset, so if a single once has no `time` dimension, it crashes.

PR #416, opened 2024-06-20
Choose which dimensions are reduced in `measures_improvement`

coxipi:
### Pull Request Checklist:
    - This PR fixes #xyz

* `measures_improvement` has an extra kwarg: `dim`, which allows to control on which dimensions we compute the percentage of improvement.

Default behaviour `None` is the same as before, reduce all dimensions. I thought it made sense to put `dim` as the first kwarg, so this is not breaking per-se, but still somewhat semi-breaking for scripts inputting `to_level` positionnally e.g. `measures_improvement([ds1,ds2], to_level)`, so I can move down `dim` if it's preferrable. 

There is a small breaking change: `ds1-ds2` is used instead of `ds2` to find non-null values and compute the pct of improvement. This may result in less non-null values. I think it's the right thing to do though.

### Other information:

My example is that I had a dataset with (rlat, rlon, period), I want improvement percentages for each period. I realize now this is an unconventional dataset in the xscen philosophy, I should have had separate dataset, when using `properties_and_measure` I should have used `period` etc etc. and have multiple catalog entries. But, for properties using  `group=time.season, time.month`, we would have an extra dimension `season, month` that we don't necessarily want to reduce, so I think it still makes sense.

Also: I think this is unrelated to templates, I assumed this is related to catalogs? So I checked the box


coxipi:
For now, I assume that if `dim` is precised, it is present in every variable. 

If we want even more control, we could accept `dim` as a dict, where every key is a given variable. I think this would not be used that often.  With `dim=None`, all dimensions are reduced like before, regardless if dimensions are different in distinct indicators (e.g.  (time) vs. (time,season))

CHANGELOG.rst
```diff
@@ -18,6 +18,7 @@ Internal changes
 * Added pins to `xarray`, `xclim`, `h5py`, and `netcdf4`. (:pull:`414`).
 * Add ``.zip`` and ``.zarr.zip`` as possible file extensions for Zarr datasets. (:pull:`426`).
 * Explicitly assign coords of multiindex in `xs.unstack_fill_nan`. (:pull:`427`).
+* `diagnostics.measures_improvement` now accepts dim`, which specifies `dimension(s)` on which the proportion of improved pixels are computed. (:pull:`416`)
```
coxipi, line 21:
```suggestion
* `diagnostics.measures_improvement` now accepts `dim`, which specifies `dimension(s)` on which the proportion of improved pixels are computed. (:pull:`416`)
```

src/xscen/diagnostics.py
```diff
@@ -536,13 +538,20 @@ def measures_improvement(
         List of 2 datasets: Initial dataset of measures and final (improved) dataset of measures.
         Both datasets must have the same variables.
         It is also possible to pass a dictionary where the values are the datasets and the key are not used.
+    dim : str or sequence of str, optional
+        Dimension(s) on which to compute the percentage of improved grid points. Default is `None`, which reduces
+        all dimensions.
```
coxipi, line 543:
```suggestion
    dim : str or sequence of str, optional
        Dimension(s) on which to compute the percentage of improved grid points. Default is `None`, which reduces all dimensions.
```

src/xscen/diagnostics.py
```diff
@@ -554,24 +563,29 @@ def measures_improvement(
...
             diff_bias = abs(ds1[var]) - abs(ds2[var])
-        diff_bias = diff_bias.values.ravel()
-        diff_bias = diff_bias[~np.isnan(diff_bias)]
+        total_improved = (diff_bias >= 0).sum(dim=dims)
+        total_notnull = diff_bias.notnull().sum(dim=dims)
+        percent_better_var = total_improved / total_notnull
+        percent_better.append(percent_better_var.expand_dims({"properties": [var]}))
 
-        total = ds2[var].values.ravel()
-        total = total[~np.isnan(total)]
```
coxipi, line 567: I realized that I used `diff_bias.notnull().sum(dim=dims)`. If `diff_bias` has more null values than `ds2[var]` (i.e.  `ds1[var]` set of coordinates with null values is not a subset of those of `ds2[var]`), then this could make a difference. 

I could write `total_notnull = ds2[var].notnull().sum(dim=dims)` although I think in the end, we might prefer `diff_bias.notnull`. Thoughts?

coxipi: I think the current definition makes more sense, but it will be breaking, so I'd like to validate the change

src/xscen/diagnostics.py
```diff
@@ -7,6 +7,7 @@
 from copy import deepcopy
 from pathlib import Path
 from types import ModuleType
+from typing import Union
```
RondeauG, line 10: Do you really need it? I think that it's mostly obsolete since Python 3.10.

RondeauG (commented):
LGTM, but I'll let @juliettelavoie approve the PR since she's a lot more familiar with this part of the code.

src/xscen/diagnostics.py
coxipi: Got it, I also changed some xscen docstrings and replaced Union with the pipe |

tests/test_diagnostics.py
```diff
@@ -448,6 +448,32 @@ def test_measures_improvement(self):
...
+            [
+                m2.quantile_98_tas.to_dataset().expand_dims({"dummy_dim": [i]})
+                for i in range(3)
+            ],
+            dim="dummy_dim",
+        )
+        out = xs.diagnostics.measures_improvement(
+            [m3, m3], dim="dummy_dim", to_level="test"
+        )
+        assert set(out.dims) == {"properties", "season"}
```
juliettelavoie, line 474: season?

juliettelavoie: je n'ai aussi pas dummy_dim quand je roule ça de mon bord..

juliettelavoie: ah non mon erreur dsl, j'utilisais le mauvais yaml

coxipi:
The current test fail is in `TestAdjust::test_basic` which should not be affected by my changes

juliettelavoie (approved):
super!
I think the breaking changes for the nan and positional `to_level` are minor enough.

FYI, The templates are here: https://github.com/Ouranosinc/xscen/tree/main/templates
the first ones uses `measures_improvement`, but your changes are not breaking enough to break that script.

PR #430, opened 2024-07-24
Unpin cartopy

aulemahal:
Cartopy was pinned in #403. I think this was because of the change of license. However, cartopy 0.23 is not available on Narval and before asking for it, I wanted to be sure I could argument for it.

After searching a bit, I think we do not need to pin as previous cartopy versions were released under the "LGPL", _not GPL_. LGPL allows dynamical linking, which is what a `import` does, to my knowledge.
See discussions :
https://opensource.stackexchange.com/questions/9470/can-i-sell-a-proprietary-software-with-an-lgpl-library-bundled-along-with-it-wi
and
https://stackoverflow.com/questions/8580223/using-python-module-on-lgpl-license-in-commercial-product


aulemahal:
Resolved on narval.

PR #442, opened 2024-08-14
Fix usage of tilde operator instead of not when saving regridder weights

bstdenis:
This fixes a bug that leads to the regridding weights being saved over and over again even when they are being reused from a file on disk.

Note that ~os.path.isfile('/etc/hosts') == -2, can't use that operator in Python for negation!


aulemahal (approved):
woupsi! Merci!

P.S. : Je t'ai ajouté au repo avec des droits d'écriture, plus besoin de faire une fourche!

PR #443, opened 2024-08-15
Take translation out of the build process

### LINKED ISSUE #342: Can't compile translation catalog when installing from source
### Setup Information

- xscen version: latest
- Python version: all
- Operating System: any

### Description

The setuptools hack we added in `setup.py` only works for `pip install .` or `pip install xscen`, but no when the argument is a git url or if the `-e` option is specified. This is gossing.

### Steps To Reproduce

```
pip install git+https://github.com/ouranosinc/xscen.git
python -c "import xscen"
```

### Additional context

Opinion : I though `gettext` and `babel` would help the work of translating xscen. I do not have that feeling anymore. For some reason, babel only canonically works with setuptools.

There would alternatives : we might be able to hack something by using a build backend that allows running custom commands during the build process (poetry seems to allow that).

OR : we go back to the simple method that xclim uses : put all the strings in a text file (yaml would be better than json). In the end, `gettext` only... gets texts. We could reimplement the core of what we use from it in only 5-6 lines. And remove the need for "compilation" completely.

### Contribution

aulemahal:
### Pull Request Checklist:
    - This PR fixes #342

* Compiled french translations are now kept within the repo and package. No need to `make translate` on each clone of the source.
* A new Makefile entry `make checkfrench` ensures that the french catalog is up-to-date AND that the compiled version is older than the catalog. See notes.
* The linting CI runs this `checkfrench` so a PR adding translatable messages cannot be merged if the french translation is not given.

No.

The only change is for developers. `make translate` needs only to be run if a translatable message was added, after `make findfrench` was run and the "xscen.po" file updated with the french translation. So, _exactly_ what the contributing guide says to do. 

### Other information:
The date check on `mo` files is weak : a catalog file could have a modified time changed even tough no real change was done. And (in theory) a compiled file could have a recent date even though it is not in sync wit the catalog. However, binary comparion of two `mo` files taken from the same catalog fails. My guess is that the compilation time is included in the mo file, and I didn't want to dig deeper. We don't update the translation catalogs often, so this should be enough.

@Zeitsperre This removes the `setup.py`  stuff. Feel free to change the build system if you feel like doing so!


aulemahal:
As seen on the failure of commit aaee623, the new CI element works as expected.

aulemahal:
And I checked: installing from this branch with "pip git+https://..." works, while it does not for main.

CHANGELOG.rst
```diff
@@ -20,6 +20,7 @@ Internal changes
 * Added pins to `xarray`, `xclim`, `h5py`, and `netcdf4`. (:pull:`414`).
 * Add ``.zip`` and ``.zarr.zip`` as possible file extensions for Zarr datasets. (:pull:`426`).
 * Explicitly assign coords of multiindex in `xs.unstack_fill_nan`. (:pull:`427`).
+* French translations are compiled offline. A new check ensures no PR are merged with missing messages. (:issue:`342`, :pull: `443`).
```
Zeitsperre, line 23:
```suggestion
* French translations are compiled offline. A new check ensures no PR are merged with missing messages. (:issue:`342`, :pull:`443`).
```

Zeitsperre (approved):
Well done!

Zeitsperre:
@aulemahal Do we want to enable the French RTD documentation in this PR as well?

aulemahal:
Je ne pense pas qu'il y ait un besoin et je ne suis personnellement pas à tout à fait à l'aise avec une traduction automatique.

PR #474, opened 2024-10-07
Add get_warming_level_from_period, rename old function

RondeauG:
### Pull Request Checklist:
    - This PR fixes #xyz

* Adds `xs.get_warming_level_from_period` to find the warming level associated to a given period.

* For clarity, I changed `get_warming_level` to `get_period_from_warming_level`.

### Other information:

* This function is needed if we want to easily implement @coxipi's modified warming levels.


src/xscen/extract.py
```diff
@@ -1068,6 +1035,174 @@ def _get_warming_level(model):
...
 
 
+@parse_config
+def get_warming_level_from_period(
+    realization: (
+        xr.Dataset | xr.DataArray | dict | pd.Series | pd.DataFrame | str | list
+    ),
+    period: list[str],
+    *,
+    tas_baseline_period: Sequence[str] | None = None,
```
coxipi, line 1045: since the first argument is `period`, perhaps the second argument should simply be `baseline_period`?

aulemahal:
> We'll also eventually need a column in our global_tas file with the observed temperatures.

Don't we already have 6 such sources ?
```
>>> ds.where(ds.experiment == 'obs', drop=True).source
<xarray.DataArray 'source' (simulation: 6)> Size: 384B
array(['Berkeley-Raw', 'Berkeley', 'GISTEMPv4', 'HadCRUT5',
       'NOAAGlobalTempv5', 'JRA-55'], dtype='<U16')
Coordinates:
    mip_era      (simulation) <U5 120B 'obs' 'obs' 'obs' 'obs' 'obs' 'obs'
    source       (simulation) <U16 384B 'Berkeley-Raw' 'Berkeley' ... 'JRA-55'
    experiment   (simulation) <U6 144B 'obs' 'obs' 'obs' 'obs' 'obs' 'obs'
    member       (simulation) <U10 240B '' '' '' '' '' ''
    data_source  (simulation) <U33 792B 'Computed following WMO guidelines' ....
Dimensions without coordinates: simulation
```

src/xscen/extract.py
RondeauG: My idea was to keep the arguments pretty much the same between `get_warming_levels` and this new function.

src/xscen/extract.py
```diff
@@ -1068,6 +1035,174 @@ def _get_warming_level(model):
...
+        ):
+            raise ValueError(
+                f"Period {period} is not fully covered by the provided 'tas_src' database for {selected}."
+            )
+
+        # compute reference temperature for the warming and difference from reference
+        wl = (
+            tas_sel.sel(time=slice(*period)).mean()
+            - tas_sel.sel(time=slice(*tas_baseline_period)).mean()
+        )
```
coxipi, line 1116:
```suggestion
def IPCC_rolling(da, window=20, dim="time"): 
    out = da.rolling({dim:window}, center=True).mean()
    if window % 2 == 0:  # Even window
        out = out.shift({dim:-1})
    return out 
        wl = (
            IPCC_rolling(tas_sel).sel(time=slice(*period))
            - tas_sel.sel(time=slice(*tas_baseline_period)).mean()
        )
```
This would give a time series of warming levels. If you want a single warming level, then you would need to give a period with a single year, e.g. ("1990", "1990") (or change the input and accept a single `str`).

src/xscen/extract.py
```diff
@@ -882,7 +884,34 @@ def search_data_catalogs(  # noqa: C901
...
+    """
+    warnings.warn(
+        "get_warming_level has been deprecated. Use get_period_from_warming_level instead.",
+        FutureWarning,
+    )
+    return get_period_from_warming_level(*args, **kwargs)
+
+
+@parse_config
+def get_period_from_warming_level(  # noqa: C901
```
juliettelavoie, line 914: don't forget to change the notebook also

src/xscen/extract.py
RondeauG: Currently, this gives a single result. Taking an example from the tests:
`xs.get_warming_level_from_period("CMIP6_CanESM5_ssp585_r1i1p1f1",  ["2013", "2032"]) = 2.005`. It's the difference between the mean temperatures for 1850-1900 and 2032-2013.

For this function, since the user already provides a period, I didn't want to impose window lengths.

RondeauG:
Do we want `get_horizon_from_warming_level` instead of `period`, to match the `return_horizon` argument?

juliettelavoie:
> Do we want `get_horizon_from_warming_level` instead of `period`, to match the `return_horizon` argument?

Horizon is a more precise word than the general period, as it refers specifically to if we are showing data by temporal horizon or warming level. I like it. But then, I would also have `get_warming_level_from_horizon` to keep the symmetry.

RondeauG:
> > Do we want `get_horizon_from_warming_level` instead of `period`, to match the `return_horizon` argument?
> 
> Horizon is a more precise word than the general period, as it refers specifically to if we are showing data by temporal horizon or warming level. I like it. But then, I would also have `get_warming_level_from_horizon` to keep the symmetry.

Done. To be fully consistent, I rename `return_horizon` with `return_period`, since this is closer to what it does.

docs/notebooks/5_warminglevels.ipynb
```diff
@@ -247,13 +273,13 @@
     "\n",
     "This option is to be used when \"scalar\" subsetting is not enough, but you want to do things differently than `produce_horizons`.\n",
     "\n",
-    "Here, we'll open all experiments into a single ensemble dataset where the `realization` dimension is constructed exactly as `get_warming_level` expects it to be. We'll also average the daily data to an annual scale."
+    "Here, we'll open all experiments into a single ensemble dataset where the `realization` dimension is constructed exactly as `get_period_from_warming_level` expects it to be. We'll also average the daily data to an annual scale."
```
RondeauG, line 276:
```suggestion
    "Here, we'll open all experiments into a single ensemble dataset where the `realization` dimension is constructed exactly as `get_horizon_from_warming_level` expects it to be. We'll also average the daily data to an annual scale."
```

aulemahal:
@lpcaron has a strong opinion against _horizon_ and I'm beginning to catch on. I think "period" is more adequate for the use we make of it.  _Horizon_ as a notion of "limit" that doesn't apply.

juliettelavoie:
> @lpcaron has a strong opinion against _horizon_ and I'm beginning to catch on. I think "period" is more adequate for the use we make of it. _Horizon_ as a notion of "limit" that doesn't apply.

not sure I get the notion of "limit" with horizon... 


Note that the IPCC atlas uses "periods" to choose between time horizon/period and GWL 
Capture d’écran, le 2024-10-07 à 15 13 06

RondeauG:
We use `horizon` in multiple places throughout `xscen`, including as a dimension/coordinate and as an argument. Does that mean that we should change all of those to `period`?

juliettelavoie:
The Copernicus Interactive Climate Atlas also uses the word near term, medium term and long term. `get_term_from_warming_level` ?

aulemahal:
We use it in xscen because Ouranos uses it everywhere...

Sorry to raise this somewhat unrelated issue on this PR. We could also merge this with whatever solution and discuss the rest in a larger group.

But, the "limit" I'm mentioning comes from the first meaning of "horizon". When used temporally, I hear it as a blurry timeframe set in the future in a way a bit slightly precise than "term". I feel that the canadians translators agree with me : https://www.btb.termiumplus.gc.ca/tpv2alpha/alpha-eng.html?lang=eng&i=1&srchtxt=Horizon&index=alt&codom2nd_wet=1#resultrecs

Like, a common construction like "Horizon 2050" means a deadline set in 2050. In climate terms, we usually assume that it means that 2050 is in the middle (ish) of the period from which we want statistics, but that's not what the term means, rather it is our scientific analysis of what data is needed.

juliettelavoie:
I think we can keep horizon in xscen but with a specific definition.

- Use `period(s)` we are cutting time to a given period but we keep the frequency (usually day). The format is ['XXXX','XXXX'].
- Use `horizon`, we have a single value for that horizon (usually a climatological mean) with a `horizon` coord that has the format "XXXX-XXXX" or even "+XdegC". Like the output of 'xs.produc_horizon`.

With this logic, we would have `get_period_from_warming_level` because the output is a list and the climatological mean has not been done yet.

RondeauG:
Right, so this is my suggestion:

- `get_warming_level` --> `get_period_from_warming_level(realization, wl, window, ..., return_period)`
- New function `get_warming_level_from_period(realization, period)`.

lpcaron:
> And I think Ouranos needs to clearly define the vocabulary for explaining the framework used to show results: GWL or time [enter consensus word here].

Agreed. I'm working hard to remove the mentions and the use of the word horizon, but there is a lot of inertia in the system...

docs/notebooks/5_warminglevels.ipynb
```diff
@@ -166,26 +166,49 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "6",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Get the warming level associated to 2041-2070\n",
+    "print(\n",
+    "    xs.get_warming_level_from_period(\n",
+    "        \"CMIP6_CanESM5_ssp585_r1i1p1f1\", horizon=[\"2041\", \"2070\"]\n",
```
RondeauG, line 203:
```suggestion
    "        \"CMIP6_CanESM5_ssp585_r1i1p1f1\", [\"2041\", \"2070\"]\n",
```

CHANGELOG.rst
```diff
@@ -9,6 +9,11 @@ Contributors to this version: Gabriel Rondeau-Genesse (:user:`RondeauG`).
 New features and enhancements
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 * ``xs.io.make_toc`` now includes the global attributes of the dataset after the information about the variables. (:pull:`473`).
+* New function ``xs.get_warming_level_from_period` to get the warming level associated with a given time horizon. (:pull:`474`).
```
RondeauG, line 12:
```suggestion
* New function ``xs.get_warming_level_from_period`` to get the warming level associated with a given time horizon. (:pull:`474`).
```

juliettelavoie (approved):
merci Gab!

PR #478, opened 2024-10-18
Skip dirs in parse_directory

aulemahal:
### Pull Request Checklist:
- [ ] This PR addresses an already opened issue (for bug fixes / features)
    - This PR fixes #xyz
- [x] (If applicable) Documentation has been added / updated (for bug fixes / features).
- [x] (If applicable) Tests have been added.
- [x] This PR does not seem to break the templates.
- [x] CHANGELOG.rst has been updated (with summary of main changes).
  - [ ] Link to issue (:issue:`number`) and pull request (:pull:`number`) has been added.

### What kind of change does this PR introduce?

* `parse_directory` : New argument `skip_dirs`, a list of folders to not go into when parsing.

### Does this PR introduce a breaking change?
No.

### Other information:
The MRCC5 catalog (him again) has two issues this help fix:
- Some simulations are copied in two sub-folders of the list of `directories` to parse. This is because I want to give some time to people to adapt to the new disk. The entries are exactly the same except for the path, so there's no way to filter them out currently with the other arguments of `parse_directory`.
- There are some many files.... This skipping will help the parsing time by skipping folders I _know_ are garbage and would be rejected anyway by other filters.

PR #479, opened 2024-10-21
Skip "root" dirs too

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

Small fix upon #478. Skips even if a directory to skip is one of the entries in `directory` or a parent of one of the entries.
I implemented it because I misunderstood an error I was having, but still, this could be used in the specific scenario where you are running `parse_directory` through an incremental stack of xscen configurations. To remove en entry of the  `directories` list of a previous configuration, you would only need to add a `skip_dirs` entry, and not rewrite the filtered  `directories`.

This also adds logging calls when skipping. And a precision to the doc about the absoluteness of paths to skip.

No.


RondeauG (approved):
LGTM.

PR #480, opened 2024-10-24
sdba Adjust methods work in adjust function

coxipi:
### Pull Request Checklist:
    - This PR fixes #xyz

* `dtrain` in ``adjust`` is optional, you can instead pass a `method` str, to use a biasadjust method that only uses adjust

Not for now. If we would like a more natural ordering or arguments in the ``adjust`` signature, then we would need some breaking changes (for now I've put the new arguments at the very end of the signature)

### Other information:

* `group` needs to be passed in `xclim_adjust_args` when using this method. It can be the unparsed group. Is there a problem with this way of doing this? IMO, this is better than having to explicitly pass `group` to the function. 
I my YAML scripts, I directly put the group in `xclim_train_args`,  and I know that those kwargs must be parsed at some point, e.g. because `group` comes from a YAML, is not yet a Grouper.


src/xscen/utils.py
```diff
@@ -867,12 +867,12 @@ def clean_up(  # noqa: C901
     variables_and_units : dict, optional
         Dictionary of variable to convert. e.g. {'tasmax': 'degC', 'pr': 'mm d-1'}
     convert_calendar_kwargs : dict, optional
-        Dictionary of arguments to feed to xarray.Dataset.convert_calendar. This will be the same for all variables.
+        Dictionary of arguments to feed to xr.Dataset.convert_calendar. This will be the same for all variables.
```
coxipi, line 870:
```suggestion
        Dictionary of arguments to feed to xarray.Dataset.convert_calendar. This will be the same for all variables.
```

RondeauG:
Is this PR still relevant? Is it ready for review? @coxipi

coxipi:
closing this for now: Too much has changed, and it's not a priority for now

PR #483, opened 2024-10-25
More flexible `produce_horizon`

RondeauG:
### Pull Request Checklist:
    - This PR fixes #xyz

* You can now skip the computation of indicators, if you've already done that step and just want the subsetting/climatological_op utilities of `produce_horizon`.
* You can now control the `op` in `produce_horizon`.
* Fixed a bug in `compute_indicators` where the `cat:variable` was determined too early and only gave the first indicator even when multiple were computed.

- `indicators` was moved to a `kwarg`, since it is now optional.

### DIFF
--- a/src/xscen/aggregate.py
+++ b/src/xscen/aggregate.py
@@ -955,7 +955,7 @@ def produce_horizon(  # noqa: C901
     ds : xr.Dataset
         Input dataset with a time dimension.
         If 'indicators' is None, the dataset should contain the precomputed indicators.
-    indicators :  Union[str, os.PathLike, Sequence[Indicator], Sequence[Tuple[str, Indicator]], ModuleType], optional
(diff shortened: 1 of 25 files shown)


src/xscen/aggregate.py
```diff
@@ -928,16 +928,18 @@ def spatial_mean(  # noqa: C901
 @parse_config
 def produce_horizon(  # noqa: C901
```
juliettelavoie, line 929: maybe add a bit more explanation in the docstring that if you dont give `indicators` , `ds` should be an indicators.

juliettelavoie: there could also be a check with a warning. like ds is daily and you didn't give `indicator`. this is could go wrong. `clim_op` wants a indicator as input.

juliettelavoie (approved):
lgtm!

src/xscen/aggregate.py
juliettelavoie: forgot about that. in that case, I don't think it's necessary

PR #492, opened 2024-11-11
Tests for the IO module

RondeauG:
### Pull Request Checklist:
    - This PR fixes #xyz

* Mainly adds tests to the IO module, along with bugfixes:
  * Adds a missing library (`openpyxl`) to the requirements.
  * Improves the code in `estimate_chunks` to support n dimensions.
  * Fixes a bug in `subset_maxsize` where the last year would be dropped.
  * Fixes a bug in `clean_incomplete` where the `zmetadata` file would not be removed.
  * Added more flexibility to `clean_incomplete`.
  * Fixed a bug in `_coerce_attrs` where encoding would get dropped.
  * Fixed bugs in `save_to_zarr` with `mode=='a'`: encoding of existing variables is now always dropped & the code now crashes if trying to both append dimensions with new data and add a new variable to an existing zarr.
  * Fixed bugs in `save_to_zarr` where the code would crash when skipping a variable if `bitround` was not None.
  * Fixed a bug in `make_toc` where the French name for the sheet would get dropped if given a Dataset instead of a DataArray.
  * Better default argument for `column` in `save_to_table`. It stays unchanged if given a Dataset with multiple variables, but is now `None` otherwise.
  * Fixed a bug in `save_to_table` where `engine_kwargs` would not be used if saving an Excel with a single sheet.

- An option was added to `publish_release_notes` to only print the notes relevant to the latest version, rather than the whole Changelog.

-To remove some clutter,  `publish_release_notes` and `show_versions` have been moved from `utils.py` to `testing.py`.


src/xscen/io.py
```diff
@@ -272,22 +248,42 @@ def clean_incomplete(path: str | os.PathLike, complete: Sequence[str]) -> None:
...
+
+        for fold in filter(lambda p: p.is_dir(), path.iterdir()):
+            if fold.name not in complete:
+                msg = f"Removing {fold} from disk"
+                logger.warning(msg)
+                sh.rmtree(fold)
+
+    elif incomplete is not None:
+        for fold in filter(lambda p: p.is_dir(), path.iterdir()):
+            if fold.name in incomplete:
```
juliettelavoie, line 279: what happens if somebody puts a coord in incomplete ? will it break the dataset ? maybe put a warning or error ?

src/xscen/io.py
```diff
@@ -272,22 +248,42 @@ def clean_incomplete(path: str | os.PathLike, complete: Sequence[str]) -> None:
...
+                msg = f"Removing {fold} from disk"
+                logger.warning(msg)
+                sh.rmtree(fold)
+
+    # Remove .zmetadata to avoid issues with zarr and xarray
+    if (path / ".zmetadata").exists():
+        Path.unlink(path / ".zmetadata")
```
juliettelavoie, line 286: when the calculation that failed is redone, will it add a new complete zmetadata ?
why don't we just regenerate a correct one here?

src/xscen/io.py
```diff
@@ -510,24 +509,33 @@ def _skip(var):
 
         if mode == "o":
             if exists:
-                var_path = path / var
-                msg = f"Removing {var_path} to overwrite."
-                logger.warning(msg)
-                sh.rmtree(var_path)
+                clean_incomplete(path, incomplete=[var])
```
juliettelavoie, line 512: pas si grave et I see that it does the same, but the name is a bit confusing. Using mode='o' is not necessarily because it was incomplete.

src/xscen/testing.py
```diff
@@ -242,3 +249,185 @@ def fake_data(
...
+        If provided, prints to the given file-like object. Otherwise, returns a string.
+    deps : list, optional
+        A list of dependencies to gather and print version information from. Otherwise, prints `xscen` dependencies.
+
+    Returns
+    -------
+    str or None
+    """
+    if deps is None:
+        deps = [
```
juliettelavoie, line 361: I'm not sure we will remember to modify this when we change deps. Is there a way to pull it from somewhere without explicetely naming them all?

tests/test_testing.py
```diff
@@ -0,0 +1,67 @@
...
+            assert len(out.split("\n\n## v0.")) == 2
+        else:
+            assert len(out.split("\n\n## v0.")) > 2
+
+
+def test_show_version(tmpdir):
+    xs.testing.show_versions(file=tmpdir / "versions.txt")
+    with Path(tmpdir).joinpath("versions.txt").open(encoding="utf-8") as f:
+        out = f.read()
+    assert "xscen" in out
```
juliettelavoie, line 60: if we have a way to access the deps without listing them as suggested above we could use it here also

juliettelavoie (approved):

src/xscen/io.py
RondeauG: Good catch. Just to be safe, we'll always skip coordinates and dimensions that are correctly detected when we open the dataset.

src/xscen/testing.py
RondeauG: I've found a way to do that (and by "I", I mean Copilot wrote it all).

PR #501, opened 2024-12-16
Strip catalog attributes - Optional reconstruction version.

### LINKED ISSUE #499: `save_to_*` should remove catalog-dependent attributes.
### Generic Issue

I think the idea of the `cat:XXXX` attrs that `intake-esm` inserts from the catalog columns was that it is an information that comes "from the catalog" and opposed to "from the data source". As such, I think the saving function should remove these attributes. Of course, we can't do anything is xarray is called directly on the dataset, but at least xscen function would be aware of the distinction and drop the stuff.

This way we don't carry over obsolete catalog attrs as much. Like, if we open one year of data, we get the `date_start/date_end` of the files. If we then subset to, say, a month and save to disk. The wrong dates are kept in the attributes. Dropping these attributes directly in the subset function would be almost impossible, but at least it should be removed when saving. 

A standalone `strip_cat_metadata` function could be added to the API.

### LINKED ISSUE #500: Official path scheme for reconstruction shouldn't require a version.
### Generic Issue

ERA5 has no version, it  thus makes no sense for us to _require_ a version in the "original-non-sim" path pattern.

aulemahal:
### Pull Request Checklist:
    - This PR fixes #499 and fixes #500.

* New `strip_cat_attrs` utility to remove those attributes added by the catalog opening functions.
    + Use this by default in `save_to_zarr` and `save_to_netcdf`.
* In `build_path`, added possibility to have aggregated folder levels with optional components.
    + Make `version` optionnal in the non-sim schemas.

* Yes. Saved files will have a few attributes removed. 
* Yes. `build_path` will not fail on a reconstruction dataset missing a version.


RondeauG:
The failing test is what we discussed the other day with the `rechunk` function. I haven't done a fix yet, so feel free to do it here if you prefer.

RondeauG (approved):

PR #504, opened 2025-01-07
Improve partition

juliettelavoie:
### Pull Request Checklist:
    - This PR fixes #xyz

* Improve `build_partition_data`
* pinned xarray because of https://github.com/pangeo-data/rechunker/issues/154 

yes, I removed the indicator_kw  argument. I realized it was too heavy to do everything in one shot. 

### Other information:

Developped for "On the importance of the reference data: Uncertainty partitioning of bias-adjusted climate simulations over Quebec" by Lavoie et al. (submitted in 2024) (presented at the last jamboree)
Code for the paper here: https://github.com/Ouranosinc/partition


tests/test_biasadjust.py
```diff
@@ -47,11 +47,10 @@ def test_basic_train(self, var, period):
 
     def test_preprocess(self):
 
-        dref360 = self.dref.convert_calendar("360_day", align_on="year")
-
+        dhist360 = self.dhist.convert_calendar("360_day", align_on="year")
```
juliettelavoie, line 50: This test failed because ref had s 720 and sim a 730 shape, but actually it should have always failed. The function assumes that the simulation will always always a smaller or equal calendar to the ref ( ref don't have 360 day cal). 
In the previous version, we were "converting" the 360 day cal of the ref to noleap, but the function doesn't add values if you don't specify it ( which we don't because we assume that we are always only removing dates). Hence, it stays 720 long and fails.

In this version, sim is 360 and ref is convert to 360 as well. Both are 720 long

juliettelavoie: The problem was in xclim :  https://github.com/Ouranosinc/xclim/pull/2038/files

I removed the test temporarily.

CHANGELOG.rst
```diff
@@ -4,11 +4,11 @@ Changelog
 
 v0.11.0 (unreleased)
 --------------------
-Contributors to this version: Gabriel Rondeau-Genesse (:user:`RondeauG`).
+Contributors to this version: Gabriel Rondeau-Genesse (:user:`RondeauG`), Juliette Lavoie (:user:`juliettelavoie`).
 
 New features and enhancements
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-* N/A
+* Improve ``xs.ensemles.build_partition_data``. (:pull:`504`).
```
RondeauG, line 11:
```suggestion
* Improve ``xs.ensembles.build_partition_data``. (:pull:`504`).
```

environment-dev.yml
```diff
@@ -30,8 +30,8 @@ dependencies:
   - shapely >=2.0
   - sparse
   - toolz
-  - xarray >=2023.11.0, !=2024.6.0
-  - xclim >=0.53.2, <0.54
+  - xarray >=2023.11.0, !=2024.6.0, <2024.10.0 #FIXME: 2024.10.0 breaks rechunker with zarr, https://github.com/pangeo-data/rechunker/issues/154
+  - xclim >=0.54, <0.55
```
RondeauG, line 34: @Zeitsperre Are we better to always only support 1 version, or could we have `xclim >=0.53.2, <0.55` for more flexibility?

src/xscen/ensembles.py
```diff
@@ -629,13 +631,150 @@ def generate_weights(  # noqa: C901
...
+def _partition_from_list(datasets, partition_dim, subset_kw, regrid_kw):
+    list_ds = []
+    # only keep attrs common to all datasets
+    common_attrs = False
+    for ds in datasets:
+        if subset_kw:
+            ds = subset(ds, **subset_kw)
+            ds = ds.drop_vars(
+                ["lat", "lon", "rlat", "rlon", "rotated_pole"], errors="ignore"
+            )
```
RondeauG, line 643: I'd use a more generic approach.

```suggestion
            gridmap = xs.spatial.get_grid_mapping(ds)
            ds = ds.drop_vars(
                [ds.cf["longitude"], ds.cf["latitude"], ds.cf.axes["X"][0], ds.cf.axes["Y"][0], gridmap], errors="ignore"
            )
```

src/xscen/ensembles.py
```diff
@@ -629,13 +631,150 @@ def generate_weights(  # noqa: C901
...
+            concat_on=partition_dim,
+            create_ensemble_on=ensemble_on_list,
+            **to_dataset_kw,
+        )
+
+        if subset_kw:
+            ds = subset(ds, **subset_kw)
+            ds = ds.drop_vars(
+                ["lat", "lon", "rlat", "rlon", "rotated_pole"], errors="ignore"
+            )
```
RondeauG, line 746: Same as above.

src/xscen/ensembles.py
```diff
@@ -644,29 +783,36 @@ def build_partition_data(
...
     Parameters
     ----------
-    datasets : dict
-        List or dictionnary of Dataset objects that will be included in the ensemble.
+    datasets : list, dict, DataCatalog
+        List or dictionnary of Datasets or DataCatalog that will be included in the ensemble.
         The datasets should include the necessary ("cat:") attributes to understand their metadata.
-        Tip: With a project catalog, you can do: `datasets = pcat.search(**search_dict).to_dataset_dict()`.
-    partition_dim : list[str]
+        Tip: A dictionnary can be created with `datasets = pcat.search(**search_dict).to_dataset_dict()`.
```
RondeauG, line 793:
```suggestion
        Tip: A dictionary can be created with `datasets = pcat.search(**search_dict).to_dataset_dict()`.
```

src/xscen/ensembles.py
```diff
@@ -644,29 +783,36 @@ def build_partition_data(
...
         The default is ['source', 'experiment', 'bias_adjust_project'].
         For source, the dimension will actually be institution_source_member.
     subset_kw : dict, optional
         Arguments to pass to `xs.spatial.subset()`.
     regrid_kw : dict, optional
         Arguments to pass to `xs.regrid_dataset()`.
-    indicators_kw : dict, optional
-        Arguments to pass to `xs.indicators.compute_indicators()`.
-        All indicators have to be for the same frequency, in order to be put on a single time axis.
+        Note thet regriding is computationnaly expensive. For large datasets,
```
RondeauG, line 807:
```suggestion
        Note that regriding is computationally expensive. For large datasets,
```

src/xscen/ensembles.py
```diff
@@ -644,29 +783,36 @@ def build_partition_data(
...
+        it might be worth it to do do regridding first, outside of this function.
```
RondeauG, line 808:
```suggestion
        it might be worth it to do the regridding first, outside of this function.
```

src/xscen/ensembles.py
```diff
@@ -644,29 +783,36 @@ def build_partition_data(
...
-    they have to be subsetted and regridded to a common grid/point.
-    Indicators can also be computed before combining the datasets.
+    they have to be subsetted and/or regridded to a common grid/point.
 
     Parameters
     ----------
-    datasets : dict
-        List or dictionnary of Dataset objects that will be included in the ensemble.
+    datasets : list, dict, DataCatalog
+        List or dictionnary of Datasets or DataCatalog that will be included in the ensemble.
```
RondeauG, line 791:
```suggestion
    datasets : list[xr.Dataset], dict[str, xr.Dataset], DataCatalog
        Either a list/dictionary of Datasets or a DataCatalog that will be included in the ensemble.
```

src/xscen/ensembles.py
```diff
@@ -682,41 +828,32 @@ def build_partition_data(
...
-
-        if indicators_kw:
-            dict_ind = compute_indicators(ds, **indicators_kw)
-            if len(dict_ind) > 1:
-                raise ValueError(
-                    f"The indicators computation should return only indicators of the same frequency.Returned frequencies: {dict_ind.keys()}"
-                )
-            else:
-                ds = list(dict_ind.values())[0]
+    if isinstance(datasets, list):
```
RondeauG, line 833: Either this should accept both a list or a dict, or the `if` entry for dicts is missing.

src/xscen/ensembles.py
```diff
@@ -682,41 +828,32 @@ def build_partition_data(
...
+        )
 
-        if "source" in partition_dim:
-            new_source = f"{ds.attrs['cat:institution']}_{ds.attrs['cat:source']}_{ds.attrs['cat:member']}"
-            ds = ds.assign_coords(source=[new_source])
-        list_ds.append(ds)
-    ens = xr.merge(list_ds)
+    else:
+        raise ValueError(
+            "datasets should be a list or a dictionary of xarray datasets or a xscen.DataCatalog"
```
RondeauG, line 843:
```suggestion
            "'datasets' should be a list/dictionary of xarray datasets or a xscen.DataCatalog"
```

src/xscen/io.py
```diff
@@ -1053,7 +1053,7 @@ def rechunk(
         raise ValueError(
             "No chunks given. Need to give at `chunks_over_var` or `chunks_over_dim`."
         )
-
+    print(ds, chunks, worker_mem, str(path_out), str(temp_store))
```
RondeauG, line 1056: `logging.info` instead of `print` ?

environment-dev.yml
Zeitsperre: We've made a fair amount of breaking changes in recently in `xclim` in order to ensure that `xscen` works properly (which is great! Fewer bugs in `xclim`). If we want to test across more versions, we should reconfigure a CI build to run against the oldest supported version of `xclim`.

RondeauG: Alright. This is not the PR for it, but I think that having a way to actually test the oldest supported version of `xclim` would be useful (including for packages further down the line like xHydro).

src/xscen/ensembles.py
juliettelavoie: regarde plus haut, je change les dict en list

src/xscen/ensembles.py
juliettelavoie: this doesnt work exactly, but i've added a workign version

Issue #515, opened 2025-01-21
[OpenSSF] Addressing Warnings


Zeitsperre:
### Generic Issue

* xscen version: v0.10.2-dev
* Python version: Any
* Operating System: All

### Description

Barring a few small issues, our progress with the OpenSSF Best Practices are looking pretty good (https://www.bestpractices.dev/en/projects/9945). 

The only significant issue that is blocking 100% ("passing" status) is the amount of warnings emitted from the testing steps. From the scoring rubric:
* The project MUST address warnings: These are the warnings identified by the implementation of the [warnings](https://www.bestpractices.dev/en/projects/9945#warnings) criterion. The project should fix warnings or mark them in the source code as false positives. Ideally there would be no warnings, but a project MAY accept some warnings (typically less than 1 warning per 100 lines or less than 10 warnings).
* It is SUGGESTED that projects be maximally strict with warnings in the software produced by the project, where practical: Some warnings cannot be effectively enabled on some projects. What is needed is evidence that the project is striving to enable warning flags where it can, so that errors are detected early.

Many of the warnings we see are different between version of Python (due to deprecations of functions between older and newer versions of libraries). Ideally, tackling these would involve addressing warnings emitted in older Python builds first (Python3.10) and then moving onto to newer ones (this is how we tackled things in `xclim`).

Eliminating warnings is not possible (nor is it the goal), but we're seeing nearly 1700 warnings emitted in the typical testing suite, which is a bit much.

PR #516, opened 2025-01-22
Address warnings

Zeitsperre:
### Pull Request Checklist:
    - This PR partially tackles #515 

* Addresses ~950 low-effort warnings:
  * 730 emitted from `xscen.adjust` (`RankWarning: Polyfit may be poorly conditioned`) 

No.

### Other information:

Many warnings are coming from deprecated usages in `intake-esm`. It would be worthwhile to see what can be done to address them from there.

### DIFF
--- a/src/xscen/catalog.py
+++ b/src/xscen/catalog.py
@@ -186,7 +186,6 @@ class DataCatalog(intake_esm.esm_datastore):
         for datecol in ["date_start", "date_end"]:
             if datecol in self.df.columns and self.df[datecol].dtype == "O":
                 # Missing values in object columns are np.nan, which numpy can't convert to datetime64 (what's up with that numpy???)
-                # FIXME: Notebooks failing due to NaN handling of infer_objects() in pandas. dropna() is might not be the right approach.
                 self.df[datecol] = (
                     self.df[datecol]
                     .infer_objects(copy=False)
--- a/src/xscen/catalog.py
+++ b/src/xscen/catalog.py
@@ -187,7 +187,10 @@ class DataCatalog(intake_esm.esm_datastore):
             if datecol in self.df.columns and self.df[datecol].dtype == "O":
                 # Missing values in object columns are np.nan, which numpy can't convert to datetime64 (what's up with that numpy???)
                 self.df[datecol] = (
-                    self.df[datecol].infer_objects(copy=False).to_numpy().astype("datetime64[ms]")
+                    self.df[datecol]
+                    .infer_objects(copy=False)
+                    .to_numpy()
+                    .astype("datetime64[ms]")
                 )
 
         if check_valid:
--- a/src/xscen/catalog.py
+++ b/src/xscen/catalog.py
@@ -186,11 +186,7 @@ class DataCatalog(intake_esm.esm_datastore):
         for datecol in ["date_start", "date_end"]:
             if datecol in self.df.columns and self.df[datecol].dtype == "O":
                 # Missing values in object columns are np.nan, which numpy can't convert to datetime64 (what's up with that numpy???)
-                self.df[datecol] = (
-                    self.df[datecol]
-                    .dropna()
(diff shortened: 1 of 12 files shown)


src/xscen/catalog.py
```diff
@@ -187,7 +187,10 @@ def __init__(
             if datecol in self.df.columns and self.df[datecol].dtype == "O":
                 # Missing values in object columns are np.nan, which numpy can't convert to datetime64 (what's up with that numpy???)
                 self.df[datecol] = (
-                    self.df[datecol].fillna("").to_numpy().astype("datetime64[ms]")
+                    self.df[datecol]
+                    .infer_objects(copy=False)
+                    .to_numpy()
+                    .astype("datetime64[ms]")
```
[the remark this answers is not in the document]
Zeitsperre: Yep, looks like it. I've added a `fixme` for right now. Will need to address this before merging.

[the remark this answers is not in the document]
Zeitsperre: Wait, is it fixed?

src/xscen/catalog.py
```diff
@@ -187,7 +187,10 @@ def __init__(
             if datecol in self.df.columns and self.df[datecol].dtype == "O":
                 # Missing values in object columns are np.nan, which numpy can't convert to datetime64 (what's up with that numpy???)
                 self.df[datecol] = (
-                    self.df[datecol].fillna("").to_numpy().astype("datetime64[ms]")
+                    self.df[datecol]
+                    .infer_objects(copy=False)
+                    .dropna()
+                    .astype("datetime64[ms]")
```
RondeauG, line 193:
```suggestion
                    self.df[datecol]
                    .infer_objects(copy=False)
                    .to_numpy()
                    .astype("datetime64[ms]")
```
I'd like a second opinion from @aulemahal, but I think that this is the right fix. We don't want to drop empty entries, as they might come from `fx` frequencies, but convert them to NaT. This seems to do the right thing.

Minimal example:
`pd.Series(np.array([5,5,None,5])).astype("object").infer_objects(copy=False).to_numpy().astype("datetime64[ms]")`
`array(['1970-01-01T00:00:00.005',  '1970-01-01T00:00:00.005',  'NaT', '1970-01-01T00:00:00.005'], dtype='datetime64[ms]')`

aulemahal: I'm not sure I understand what `infer_objects` is supposed to do, but @Zeitsperre's solution works as the index is preserved on the series even after dropping the NaNs. And pandas understand that it must put `NaT` when merging it back.

```suggestion
                    self.df[datecol]
                    .dropna()
                    .astype("datetime64[ms]")
```

Seems that the `to_numpy` is not required anymore but don't know when it changed (I am using pandas 2.2.3).  Also, I didn't get any warnings in my tests with the previous method, do you remember what warnings we are addressing here ?

Zeitsperre:
Warnings are now below 800 in total, with the majority (600+) stemming from `intake-esm`. This is good enough for the time being, IMO.

src/xscen/catalog.py
RondeauG: There is no warning when no data is missing, but I get the warning with:

`pd.Series(np.array([5,5,5])).astype("object").fillna("").to_numpy().astype("datetime64[ms]")`

`FutureWarning: Downcasting object dtype arrays on .fillna, .ffill, .bfill is deprecated and will change in a future version. Call result.infer_objects(copy=False) instead. To opt-in to the future behavior, set `pd.set_option('future.no_silent_downcasting', True)`
`

RondeauG (approved):

src/xscen/catalog.py
aulemahal: I didn't get the warning with missing data. In a catalog's csv, the `date_start` and `date_end` column are expected to be strings or "nan", which in theory would never trigger a "downcasting". But anyway, that step seems unnecessary now.

PR #552, opened 2025-03-13
Use xclim `convert_units_to` in ``xs.train``  and ``xs.adjust``

coxipi:
### Pull Request Checklist:
    - This PR fixes #xyz

* Replace the `convert_units_to` implementation of `xsdba` by the one in `xclim` that allows to use `context="infer"`. Even if we fix the `xc.core.settings.context` thing, I think this solution might be preferred, as I understood in the discussions.

No

### Other information:

I like this method because I was a bit puzzled regarding how I would achieve automatic unit conversion with MBCn, but with this, this is almost free, I just have to revert some inconsequential changes in xsdba

### DIFF
--- a/src/xscen/biasadjust.py
+++ b/src/xscen/biasadjust.py
@@ -2,8 +2,8 @@
 
 import logging
 import warnings
+from contextlib import contextmanager
 from copy import deepcopy
-from unittest.mock import patch
 
 import xarray as xr
 import xclim as xc
@@ -56,6 +56,19 @@ def _add_preprocessing_attr(scen, train_kwargs):
 
 def _convert_units_to_infer(source, target):
     return xc.core.units.convert_units_to(source, target, context="infer")
+
+
+@contextmanager
+def xclim_convert_units_to():
+    module = xsdba.units
+    attr_name = "convert_units_to"
+    new_value = _convert_units_to_infer
+    original_value = getattr(module, attr_name)
+    setattr(module, attr_name, new_value)
+    try:
+        yield
+    finally:
+        setattr(module, attr_name, original_value)
 
 
 @parse_config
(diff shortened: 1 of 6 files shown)


src/xscen/biasadjust.py
```diff
@@ -3,11 +3,11 @@
 import logging
 import warnings
 from copy import deepcopy
+from unittest.mock import patch
```
Zeitsperre, line 6: Interesting approach. I've been reading a bit about this, and it seems like the `patch` function for `unittest.mock` is meant to be used for temporarily overriding the behaviour of a function. They suggest using it as a decorator or within a context manager *specifically for testing*, i.e.

```python
from unittest.mock import patch
import module

def new_func(*args, **kwargs):
    return "new"

with patch("module.function", new_func):
    assert module.function() == "new"
```

It might be a better idea to create a `contextmanager` for this:

```python
@contextmanager
def override(convert_func):
    original_function = module.function
    try:
        module.function = convert_func
        yield
    finally:
        module.function = original_function

```

Zeitsperre (changes requested):
The `unittest` library is made for testing. I think we should be clearer in the code that we are temporarily patching the library using a `contextmanager`.

Zeitsperre (approved):
LGTM!

tests/test_biasadjust.py
```diff
@@ -314,8 +316,8 @@ def test_xsdba_vs_xscen(
             dhistx = dhist.sel(time=slice("2001", "2003")).convert_calendar("noleap")
             dsimx = dsim.sel(time=slice("2001", "2006")).convert_calendar("noleap")
 
-            # xsdba is now climate agnostics, hydro context must be specified
-            with xc.core.units.units.context("hydro"):
+            # xsdba is now climate agnostics, patch xclim's converter
```
Zeitsperre, line 319:
```suggestion
            # xsdba is now climate-agnostic, patch xclim's converter
```

PR #557, opened 2025-03-14
Add dependency checks, move dependencies around

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #xyz

* Implements dependency checks to ensure that all declared deps are listed in the environment
* Removes several redundant sphinx extensions
* Updates the Makefile to add more defaults checks

Technically, yes. Dependencies are changed.


aulemahal:
I would argue that this line : 
https://github.com/Ouranosinc/xscen/blob/5dc77127d591b261a35c066828004997790d51db/src/xscen/catalog.py#L134
explicitly requires `pyarrow`!

environment-dev.yml
```diff
@@ -38,13 +39,15 @@ dependencies:
   - xsdba >=0.3.0
   # Opt
   - nc-time-axis >=1.3.1
-  - pyarrow >=10.0.1
+  - openpyxl
```
RondeauG, line 42: Why put `openpyxl` in optional dependencies? Is there an issue with it?

Zeitsperre: Not really. It's more that it's a library used specifically for Excel support used in one function. If a user only cares about writing to CSV or ODS, it's irrelevant.

I can mark it mandatory again, no issue.

RondeauG: Until it becomes problematic, I'd keep it as mandatory.

RondeauG (approved):

RondeauG:
I see that there are still some small discrepancies between files, such as `pyarrow`. I'll fix that then merge.

PR #594, opened 2025-05-16
Iterative creeping - save and load weights

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

* New argument `steps` to `creep_weights` so that we creep iteratively this number of time, so filling a boundary of thickness `steps * n`. The `creep_fill` detects it and applies the weights in sequence too.
* New methods to save those weights to disk and then load them back. Similar to what `xESMF` does, but for our own `creep_fill`. This syntax is completely xscen-specific. In theory CF conventions have what it needs to encode/decode sparse arrays, but the implementation is still to be done.

No.

### Other information:
I already use that technique in PC for the sea masking. And I now want to use that for SPS. The idea is that `creep_weights` is slow but `creep_fill` is fast, so if the mask is constant, you don't want to recompute it often.

Peut-être que ceci pourrait être utile à @DoMatte.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,15 @@
 Changelog
 =========
 
+v0.13 (unreleased)
+------------------
+Contributors to this version: Pascal Bourgault (:user:`aulemahal`).
+
+New features and enhancements
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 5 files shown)


RondeauG (approved):
LGTM, but I don't really know that function.

PR #609, opened 2025-06-17
add clip to additive_space

juliettelavoie:
### Pull Request Checklist:
    - This PR fixes https://github.com/Ouranosinc/xsdba/issues/164

* Add possibility to clip to nearest float before doing logit transformation

no

### Other information:
There are some 100% hurs in CMIP data.


coxipi:
Actually let me phone a friend on this one, I just want to make sure. 


@aulemahal Some CMIP6 sims have hurs=100%. Do you think applying  a `hurs.clip(None, 100^{-})`, where 100^{-} is the largest possible float{32,64} value that is below 100. So we effectively avoid problems of 100% while avoiding to define a threshold that we would need if we were jittering. But that just makes think: why jitter in the first place then e.g. for precipitations.

juliettelavoie:
Je pense qu'il est en vacances jusqu'à la st-jean, mais on peut attendre.

A difference between 0 mm/d (pr) and 100% (hurs)  is that I expect a lot more 0 pr. Jitter might be useful if it covers many quantiles, which I'm pretty sure 100 hurs won't. Just a guess though.

Another thing is maybe we should fix the dtype to float 64. Imagine a scenario with sim in float64 and ref in float32, the hurslogit of 100% hurs after clip and transform would be ~36 and ~16, respectively. So, we would learn a bias that is not really there.

PR #611, opened 2025-06-19
add standard_name to dtr def

juliettelavoie:
### Pull Request Checklist:
    - This PR fixes #xyz

* add standard_name to dtr defintion in conversion. This is to match the definition in variables.yml of xclim. The cfchecks in `health_checks` fail now because the standard name doesn't exist for dtr.

no


coxipi (approved):
Sounds good. It is a hack though, right? There is no true standard name for the diurnal temperature range, AFAIK. So if  dtr ends up in the part of the workflow where a healthcheck is performed, perhaps the health check should be less strict if we intend to be able to use non cf-standard variables?

EDIT: Ah my bad, I was checking an old version of xclim. It seems dtr has been added. Oh well, that's a xclim convention now, so at least we,re consistent

juliettelavoie:
yes, the goal here is just to be consistent with xclim

PR #616, opened 2025-07-05
Fix hurs

juliettelavoie:
### Pull Request Checklist:
- [x] This PR addresses an already opened issue (for bug fixes / features)
    - This PR fixes #xyz
- [x] (If applicable) Documentation has been added / updated (for bug fixes / features).
- [x] (If applicable) Tests have been added.
- [x] This PR does not seem to break the templates.
- [x] CHANGELOG.rst has been updated (with summary of main changes).
  - [x] Link to issue (:issue:`number`) and pull request (:pull:`number`) has been added.

### What kind of change does this PR introduce?

* Clip invalid values for hurs_from_dewpoint.
* See my issue here: https://github.com/Ouranosinc/xclim/issues/2202 

### Does this PR introduce a breaking change?
yes, previous behavior was to put nan.

### Other information:
https://github.com/Ouranosinc/xclim/pull/2203 needed for this to work.

PR #619, opened 2025-07-15
Obs utils

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

Adds the  following functions to `xs.spatial`

- `dataset_extent` which returns a "region" dict with the extent of a given dataset.
- `merge_duplicated_stations`  : merges stations with the same location from a station-obs dataset
- `voronoi_weights` : For a stations dataset, return a weight that is inversely proportional to the station density around each station. The idea is to remove station density bias when taking means of parameters computed over stations.

Also, these changes:
- Fix for scipy 1.16. This fix is different than what @juliettelavoie did in #616. Here, I preserve the behaviour. Which do we prefer ?
- Fixes in `subset`:
    + `method='gridpoint'` : when the lon/lat is a DataArray, its dimension name is preserved (instead of being replaced by `site`)
    + `method='shape'` : a shapely polygon is also accepted
- `get_grid_mapping` : allow a grid mapping to be in the data vars as well
- Doc of regions : I tried to add a `Region` typevar in the aim of having better annotations for this construct, instead of "dict2.

No.

### Other information:
I had ambitions of relying more on cf-xarray for grid mapping handling but it turns out it didn't work. Oh well.

### DIFF
--- a/src/xscen/spatial.py
+++ b/src/xscen/spatial.py
@@ -11,11 +11,13 @@ import clisops.core.subset
 import dask
 import geopandas as gpd
 import numpy as np
+import shapely as shp
 import sparse as sp
 import xarray as xr
 import xclim as xc
 from pyproj.crs import CRS
 
+from ._types import Region
 from .config import parse_config
 
 logger = logging.getLogger(__name__)
@@ -23,6 +25,7 @@ logger = logging.getLogger(__name__)
 __all__ = [
     "creep_fill",
     "creep_weights",
+    "dataset_extent",
     "get_grid_mapping",
     "subset",
 ]
@@ -242,7 +245,17 @@ def _subset_gridpoint(
     if not hasattr(lat, "__iter__"):
         lat = [lat]
 
+    dim = None
+    if isinstance(lon, xr.DataArray):
+        dim = lon.dims[0]
+        lon = lon.rename({dim: "site"})
+        lat = lat.rename({dim: "site"})
+
     ds_subset = clisops.core.subset_gridpoint(ds, lon=lon, lat=lat, **kwargs)
+
+    if dim is not None:
+        ds_subset = ds_subset.rename(site=dim)
+
     new_history = (
         f"[{datetime.datetime.now().strftime('%Y-%m-%d %H:%M:%S')}] "
         f"gridpoint spatial subsetting on {len(lon)} coordinates - clisops v{clisops.__version__}"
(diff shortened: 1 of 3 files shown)


src/xscen/spatial.py
```diff
@@ -476,3 +493,150 @@ def update_history_and_name(ds_subset, new_history, name):
...
+                    lambda x, d: (x.bfill(d) if x.dtype == "f" else x).isel({d: [0]}),
+                    args=(dim,),
+                )
+            )
+
+    h = f"[{datetime.datetime.now().strftime('%Y-%m-%d %H:%M:%S')}] Merged {N} co-located points along dimension {dim}."
+    return update_history_and_name(xr.concat(dss, dim), h)
+
+
+def voronoi_weights(
```
juliettelavoie, line 598: ça serait cool d'avoir un exemple complet de comment utiliser ça !

src/xscen/aggregate.py
```diff
@@ -261,10 +262,10 @@ def _ulinregress(x, y, **kwargs):
...
                             reg.slope,
                             reg.intercept,
-                            reg.rvalue,
-                            reg.pvalue,
-                            reg.stderr,
-                            reg.intercept_stderr,
+                            reg.rvalue if reg.slope != 0 else 0,
+                            reg.pvalue if reg.slope != 0 else 1,
+                            reg.stderr if reg.slope != 0 else 0,
+                            reg.intercept_stderr if reg.slope != 0 else 0,
```
aulemahal, line 268: @juliettelavoie  le fix est ici.

Ma logique était que "NaN" pourrait être confondu pour "données manquantes", ce qui n'est pas nécessairement le cas ?
Le cas ou tout == 0 ne devrait pas arriver souvent...

src/xscen/aggregate.py
```diff
@@ -584,7 +584,7 @@ def spatial_mean(  # noqa: C901
     method: str,
     *,
     spatial_subset: bool | None = None,
-    region: dict | str | None = None,
+    region: Region | Literal["global"] | None = None,
```
juliettelavoie: just to be sure, in reality you still just pass a dict ?

src/xscen/spatial.py
```diff
@@ -42,6 +42,7 @@
 logger = logging.getLogger(__name__)
 
 __all__ = [
+    "Region",
```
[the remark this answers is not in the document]
Zeitsperre: No problem. Python magic! :mage:

PR #636, opened 2025-08-29
Update to future intake-esm with polars

### LINKED ISSUE #618: Intake-esm polars' implementation incompatible with xscen's CSVs
Since last April, `intake-esm` has implemented reading csv files with `polars`. Which is pretty cool and should help for large ones like MRCC-disponible. However, `polars.scan_csv` does not have the same arguments as `pd.read_csv` and our built-in `read_csv_kwargs` break the new version.

This means the following

### "variable" column

We use the read kwargs to convert variable tuple reps (i.e. `"('pr',)"`) to actual tuples. `intake-esm` already has a trick for that in the `columns_with_iterables=["variable"]` init arg, which we could use. However, the `polars` implementation expects _lists_ instead of tuples. So `['pr']` instead of `"('pr',)"`. Which means we need to reformat _all_ previously existing catalog before fixing this and upgrading `intake-esm`.

Sadly, this has the consequence of storing the variables in lists, which are not hashable. This trigger errors in the test suite, where we look for "duplicates" for example (`search_data_catalogs`). I am looking into that.

### xrfreq
The read kwargs also convert old `xrfreq` to the "new" ones. 

I say that time has passed and we can simply drop that. "old" xrfreq will trigger errors and that's ok.

 
### dtypes
The kwargs ensure `string[pyarrow]` as the dtype for the "path" column and "category" for the other string columns. We could change the way we specify those to match what polars want or simply test the performance without. Maybe this is not needed anymore.

aulemahal:
### Pull Request Checklist:
    - This PR fixes #618

The newer intake-esm now uses `polars` to read catalog csv. This drops the support for some of the `read_csv` arguments we were using to enhance performance. While the current intake-esm doesn't fully support what we need, intake/intake-esm#734 fixes that. So the next release should be compatible, with the changes made here.

I also added `intake-esm`'s main in the upstream CI.

Yes, as the read_csv "converters" are not usable anymore, I dropped the `ensure_new_xrfreq` one that was executed upon the csv read. Which means catalogs using the old xrfreq specs (3H, H, etc) might break further down the line. This could added back by calling it after reading, but it felt unnecessary.

The dtypes of the columns have changed. We can force "categorical" anymore, so everything is a string, except "variable" (an object) and the dates (the way we convert those hasn't changed). I yet have to check if performance is affected. 

### Other information:
As said above, I haven't compared the performance yet.


juliettelavoie:
Merci Pascal pour les tests! J'ai vu que le gars de intake semble dire qu'on pourra choisir notre backend finalement sur ton issue. J'imagine qu'on peut juste faire ça en attendant que polars soit meilleur en recherche ?

.github/workflows/main.yml
```diff
@@ -73,10 +73,6 @@ jobs:
             python-version: "3.12"
           - os: "windows-latest"
             python-version: "3.11"
-          - os: 'ubuntu-latest'
-            python-version: "3.10"
-            esmf-version:  "8.7.0"  # 8.8.0 breaks h5py
-            xclim-version: "0.56"  # minimum supported version
```
Zeitsperre, line 79: It might be a good idea to specify at least one build that checks against the minimum support `xclim`.

tox.toml
Zeitsperre: Don't forget to modify this as well

aulemahal:
@Zeitsperre @juliettelavoie @RondeauG  This PR is now in a mergeable state.

## Zipped Zarrs
I dropped the idea of having a fix for zipped zarr in intake-esm. My PR has been sitting (with a few others) for a while now, I think the core team has had reduced time for maintenance in the last months. Instead of waiting for pushing what feels like a hack anyway, I hacked it here.

In theory, all the modifications below could be removed when the "url pipeline" stuff is implemented in zarr.

`xscen.zarrzip` exposes a new xarray engine that inherits from the zarr one, but wraps zip files in a `ZipStore`, that's all. This means that if xscen is installed, `xr.open_dataset("file.zarr.zip")` will open normally with zarr 3. However, `xr.open_zarr` **can not** be used as this function hardcodes using the default backend. This new "engine" has been added to `xs.io.get_engine`.

Also this means that opening a file with xarray now requires importing xscen because we have all modules imported directly from the top level. If xscen was previously unimported, this means the first call to `xr.open_dataset` takes a few seconds.

Intake-esm has it's own way of inferring the engine, mostly using the "format" column. Without a PR upstream, I had to hack something again. Here, I suggest skipping intake-esm's inference, letting xarray do it. This is done by passing `engine=None` to the xarray open kwargs (intake-esm will only set this if absent). I added this to the default kwargs we pass through `extract_dataset` and `Catalog.to_dataset`.   This means that this will **not** work with `Catalog.to_dataset_dict` and `Catalog.to_dask`.

As a side effect, this allows the use of xarray's `netcdf_engine_order` option by default. Previously, netcdf files where send to `netCDF4` by intake-esm when `engine` was not set. It also removes a lot of the smart code dealing with some remote storage options that intake-esm has. As I wrote in the changelog, I assumed xscen does not aim to support the same range of remote storage options. It is first and foremost aimed to deal with local files, and maybe whatever PAVICS exposes (when w get to the point of having catalogs on there).

## Catalog creation changes
With the switch to polars by intake-esm, I had to change the way we read in some columns. 

Datetime columns can be read by polars, so I changed the default kwargs. I renamed the argument to `read_kwargs`, the same as in intake-esm. In theory this is backend-agnostic, but of course pandas and polars don't have the same arguments names, so the current default expect polars to be used.

Previously, we were parsing the variable column into tuples through some special options to `pd.read_csv`. intake-esm exposed an option for that, so I made use of it (`columns_with_iterables`). Currently, this needs to be given explicitly to intake-esm, which is not great as it depends on the catalog. I suggested to add a field to the ESM collection spec, but this PR also has been sitting unanswered for a while. I implemented the idea here nonetheless. In theory, this could allow for more columns to be iterables.

In any case, I made it now clearer that xscen expects the "variable" column to be filled with tuples. `xs.parse_directory` will make sure it's the case.

Finally, now that reading catalog files is a bit more complicated, I decided to drop the possibility of passing multiples csv files to `Catalog.from_df`. No test were testing this, I am guessing it's not much used.

## Zarr 3
The new intake-esm depends on zarr >=3, so we now do too.  We had a few spots where we were dealing with zarr files manually, which I adapted or moved to use `zarr` instead.

Zarr 3 will read zarr 2 files transparently, but will write files that zarr 2 can't read. Pass `zarr_kwargs={'zarr_format': 2}` to write compatible file in `save_to_zarr`. 

Rechunker does not support zarr 3. I removed the dependency and functions in this PR, but #671 is still open to explore the alternative.  The other way is to take possession of the repo and maintain it.

## Conclusion
I expect this PR to be breaking in places I didn't think of. I think in general  reading (our current catalogs, files) will not be affected, but it is in the writing aspect I worry.

aulemahal:
Désolé tous. Ma solution d'un engine xarray custom ne fonctionne pas à tout coup avec la gestion de format/engine que fait intake-esm, du moins pas aussi facilement que je pensais.

aulemahal:
Le dernier commit enlève le backend `zarrzip` et fait du patchage tout simplement.

Ça va probablement casser un jour, mais là ça fonctionne.

PR #645, opened 2025-09-15
add clip_var

juliettelavoie:
### Pull Request Checklist:
    - This PR fixes #xyz

* Add ``clip_var`` option in ``xs.clean_up`` to clip values outside a given range.

### DIFF
--- a/.cruft.json
+++ b/.cruft.json
@@ -11,7 +11,7 @@
       "project_slug": "xscen",
       "project_short_description": "A climate change scenario-building analysis framework, built with xclim/xarray.",
       "pypi_username": "RondeauG",
-      "version": "0.13.1-dev.0",
+      "version": "0.13.1-dev.1",
       "use_pytest": "y",
       "use_black": "y",
       "use_conda": "y",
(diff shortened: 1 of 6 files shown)


src/xscen/utils.py
```diff
@@ -962,7 +966,7 @@ def clean_up(  # noqa: C901
 
     See Also
     --------
-    xclim.core.calendar.convert_calendar
+    xclim.core.calendar.convert_calendar, xarray.DataArray.clip
```
coxipi, line 969: I'm not sure if base operations in xarray should be included in "See Also"? Otherwise, you might also want to add e.g. xr.DataArray.round, etc, and this never stops? I don't know, I also understand that 'clip' is less transparent than 'round', so maybe this justifies its inclusion. Anyways, feel free to ignore this nitpick

coxipi (approved):
Looks good!

src/xscen/utils.py
juliettelavoie: I think it should include all function where xs.clean_up directly passes the args to the external function.

aulemahal (approved):

PR #661, opened 2025-11-03
Fix correct time

### LINKED ISSUE #660: ensure_correct_time fails if two variables have different times
### Setup Information

- xscen version: 0.13
- Python version: 3.11.5
- Operating System: linux

### Description

If tasmax has time at 00:00, but hurs has time at 12:00, and I try to open both, `ensure_correct_time` fails.

If we had passed them individually, both would have been floored and there would be no issue. I think xscen should be able to handle both at the same time.. or do we want this to fail ?

### Steps To Reproduce

```
d = xs.search_data_catalogs(variables_and_freqs= {'tasmax': 'D', 'hurs': 'D'},
                             other_search_criteria= {'id': 'CMIP6_ScenarioMIP_INM_INM-CM4-8_ssp245_r1i1p1f1_global'},
                            data_catalogs= ['simulation.json'],
                           ).popitem()[1]
ds_sim = xs.extract_dataset(catalog=d)
```
->
```
ValueError: Dataset is labelled as having a sampling frequency of D, but some periods have more than one data point.
```

### Additional context

_No response_

### Contribution

juliettelavoie:
### Pull Request Checklist:
    - This PR fixes #660 

* Floors data in the preprocess.

won't fail with two variables with different times, but should be the same for the rest

### Other information:

I left the ``ensure_correct_time`` chexk where it was to make sure it is okay with the combined dataset.
Also, I wanted to avoid changing `ensure_correct_time` to avoid a breaking change if somebody was using that function directly. 

EDIT: Je ne comprends pas l'erreur dans le tests Lint. Please help!


juliettelavoie:
argh! I forgot that we can't extract the pangeo data. and the data in samplecat is monthly.
I assume this was because daily data was too big ?

not sure how to test this PR without daily data..

PR #665, opened 2025-11-21
Reduce warnings

RondeauG:
### Pull Request Checklist:
    - This PR fixes #xyz

* Solves a few warnings primarily coming from `xarray`.

* dtype encodings are now removed before saving, since it could create inconsistencies. It can be accomplished through kwargs instead,
  * For example, the `air` variable of the xarray tutorial dataset has a `int64` in its encoding, when it's clearly a float and recognized as such by xarray.

### Other information:

* I'm not sure what the right fix is for `pd.concat([dc.df for dc in data_catalogs], ignore_index=True)`. The warning was that in the future, NaNs would be taken into account when determining the dtype of a column. In our case, this can happen if we open `simulation` and `reconstruction` catalogs. I tried to put the right dtypes, but then I had issues with subsequent tests, so I've left it as `object` mostly everywhere.
* For the various `concat`/ `merge`, I've set them as the current defaults.

### DIFF
--- a/src/xscen/extract.py
+++ b/src/xscen/extract.py
@@ -605,29 +605,17 @@ def search_data_catalogs(  # noqa: C901
         cat_kwargs = {"registry": registry_from_module(load_xclim_module(conversion_yaml))}
 
     # Prepare a unique catalog to search from, with the DerivedCat added if required
-    all_columns = set()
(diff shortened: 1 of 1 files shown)


juliettelavoie (approved):
lgtm

PR #692, opened 2026-02-11
Support Pandas 3, Python 3.14, unpin numcodecs

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

* Remove pin for Pandas 3
    * Small fix in `DataCatalog.__init__` because strings now have their own dtype
    * Small fix in `subset_warming_level` because `get_period_from_warming_level` might return nan and `slice(nan)` raises. I'm not sure where the change is here. 
    * Fix a deprecation warning for `pd.Timedelta(1, 'd')` (only `D` should be used)
* Support python 3.14
    * This change required the change in `test_spatial`, I think because it allowed something concerning xarray ? Anyway, the error raised by the test changed.
    * @Zeitsperre  Have I done all things needed for a python bump ?
* Remove pin for `numcodecs`. I don't remember the issue, but the version installable now works well.
* Adapt a few comments about pins and errors.

No.

### Other information:
Locally, I test against `xsdba` 0.6.0, so this seems supported as well.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,7 @@ Contributors: Juliette Lavoie (:user:`juliettelavoie`), Gabriel Rondeau-Genesse 
 
 New features and enhancements
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+* xscen now supports numcodecs > 0.16, Pandas 3 and Python 3.14 (:pull:`692`).
 * The `group=False` option was added to ``xs.train`` to completely skip this argument when calling `xsdba`. (:pull:`668`).
 * The `stack_periods` argument was added to ``xs.adjust`` to pass arguments to ``xsdba.stack_periods`` before adjustment. (:pull:`668`).
 * A conversion from geopotential to orography has been added to the xclim modules. (:pull:`673`).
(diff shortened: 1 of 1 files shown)


environment-dev.yml
```diff
@@ -20,10 +20,9 @@ dependencies:
   - intake-esm ==2025.2.3
   - matplotlib >=3.6
   - netcdf4
-  - numcodecs <0.16.0
```
Zeitsperre: This is explicitly imported, so it needs to come back.

environment.yml
```diff
@@ -22,10 +22,9 @@ dependencies:
   - intake-esm ==2025.2.3
   - matplotlib >=3.6
   - netcdf4
-  - numcodecs <0.16.0
```
Zeitsperre: Re-add this without the pin and we're Gucci.

pyproject.toml
Zeitsperre: You can also repin these at the same time:
```suggestion
"nbsphinx >=0.9.8",
"nbval",
"sphinx >=8.2.0",
```

aulemahal: This makes the CI fail because sphinx 8.2 doesn't support python 3.10 and 3.11.

Should we drop 3.10 and 3.11 or remove doc deps from `environment.yml` (and use pip's groups for RTD maybe?)

The other failure is about pydantic and rust not being able to be compiled on python 3.14. Doesn't happen on conda on my machine.

Zeitsperre: I would just pin the python version in the docs requirements higher.

The `pydantic` error speaks to me for some reason. I'll look into that tomorrow.

Zeitsperre: Gotcha. This is why I typically split the `docs` requirements into their own file. Will take a stab at this later today.

PR #703, opened 2026-03-27
Fix Pandas 3 change when reading list of variables from files after parsing variables as strings

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

This is a pattern used for MRCC monthly files. The file name contains a variable name, but the file itself contains more than that. Thus, "variable" appears in the pattern and is parsed, but to have all variables, we read them from file with this argument : `read_from_file=[['variable', ...], ['variable']]`. 

Since Pandas 3, a new StringDType exists, for the result of the first filenames parsing has a `variable` column of this dtype. And when setting an element to a tuple, it fails. Earlier, it was an Object dtype, so there were no issue. 

The fix simply changes the dtype of the column on the fly when this situation happens.

No.


Zeitsperre (approved):
Not sure if I follow, a few questions:

- What does it look like to adapt our database/naming conventions to these changes?
- Are these changes backwards-compatible?

PR #718, opened 2026-04-17
fix spatial mean

### LINKED ISSUE #717: spatial_mean doesn't work with CRCM5
### Setup Information

- xscen version: 0.14
- Python version: 3.12
- Operating System: linux

### Description

In spatial_mean, we only create bounds when we find `rotated_pole` (not any coords that might have the same info, like `crs`). https://github.com/Ouranosinc/xscen/blob/67350ed2e4a277e6aae4e9df15880bd70811e562/src/xscen/aggregate.py#L685

This means that spatial_mean doesn't work on CRCM5 data, if you don't load the bounds.

ex.

```
cats= xs.DataCatalog('simulation.json')
ds1=cats.search(id='CMIP6_CORDEX_CanESM5_r1i1p2f1_OURANOS_CRCM5-SN_ssp585_r1_NAM-12', xrfreq='D', variable='pr').to_dataset()
#ds1=ds1.rename({'crs':'rotated_pole'}) # uncomment to make it work
#ds1['pr'].attrs['grid_mapping']='rotated_pole'
display(ds1)
out1= xs.spatial_mean(ds1, method='xesmf',
                      region=dict(name='tmp', method='bbox',lat_bnds=[40,45], lon_bnds=[-75,-70]),
                      kwargs={'skipna':True})
```

### Steps To Reproduce

_No response_

### Additional context

it also doesn't work if we pass a DataArray instead of a Dataset. The docstring only say dataset but it is easy to forget and get a weird error.  I'll add add a warning for that also.

### Contribution

juliettelavoie:
### Pull Request Checklist:
    - This PR fixes #717 

* Add a more general check for rotated pole in spatial mean

### DIFF
--- a/src/xscen/aggregate.py
+++ b/src/xscen/aggregate.py
@@ -548,6 +548,8 @@ def spatial_mean(  # noqa: C901
     --------
     xesmf.SpatialAverager
     """
+    if isinstance(ds, xr.DataArray):
+        warnings.warn("Input is a DataArray, but should be a Dataset. This could lead to errors, especially with rotated poles.", stacklevel=2)
     kwargs = (kwargs or {}).copy()
 
     # Determine the coordinates
@@ -682,7 +684,7 @@ def spatial_mean(  # noqa: C901
         # Preemptive segmentization. Same threshold as xESMF, but there isn't strong analysis behind this choice
         geoms = shapely.segmentize(polygon.geometry, 1)
 
-        if ds.cf["longitude"].ndim == 2 and "longitude" not in ds.cf.bounds and "rotated_pole" in ds:
+        if ds.cf["longitude"].ndim == 2 and "longitude" not in ds.cf.bounds and "rotated_latitude_longitude" in ds.cf.grid_mapping_names:
             from .regrid import create_bounds_gridmapping
 
             ds = ds.assign_coords(**create_bounds_gridmapping(ds))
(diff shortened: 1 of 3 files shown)


src/xscen/aggregate.py
```diff
@@ -682,7 +684,7 @@ def spatial_mean(  # noqa: C901
         # Preemptive segmentization. Same threshold as xESMF, but there isn't strong analysis behind this choice
         geoms = shapely.segmentize(polygon.geometry, 1)
 
-        if ds.cf["longitude"].ndim == 2 and "longitude" not in ds.cf.bounds and "rotated_pole" in ds:
+        if ds.cf["longitude"].ndim == 2 and "longitude" not in ds.cf.bounds and "rotated_latitude_longitude" in ds.cf.grid_mapping_names:
```
aulemahal: In theory, we also support `oblique_mercator` and `lambert_conformal_conic`. If you
```suggestion
        if ds.cf["longitude"].ndim == 2 and "longitude" not in ds.cf.bounds:
```
then all curvilinear datasets without bounds will go through here, which is necessary anyway no ?

Then instead of xESMF raising an error on absent bounds, it will be xscen with `"Grid mapping XYZ not implemented."` or with some other error if no grid mapping is defined. Does that make sense ?

juliettelavoie: ah ok I didn't realize i would work with others, but yes sure let's have it more general

aulemahal (approved):

PR #721, opened 2026-04-28
Add theilslopes

### LINKED ISSUE #720: Add Mann–Kendall trend test option to aggregate.climatological_op
### Addressing a Problem?

`xscen.aggregate.climatological_op` currently supports the `linregress` option for linear trends. In many applications though linear regression is not always ideal especially for variables such as precipitation, where the data may contain outliers or where a non-parametric method is more appropriate.

It would be useful to add support for a Mann–Kendall test and Theil-Sen slope trend option. The Theil–Sen slope estimator gives the magnitude of the monotonic trend and the Mann–Kendall test provides a non-parametric significance test for monotonic trends. One possible implementation way would be to use `pyMannKendall`, which already provides Mann–Kendall significance tests with Theil-Sen slope estimators (although this would add a new dependency).

### Potential Solution

Add a new option `op="mannkendall"` in `aggregate.climatological_op`.

The output could follow the same structure as the existing `linregress` option, with a new dimension containing coordinates such as [`slope`, `intercept`,  `pvalue`, `Tau`, `z`].

### Additional context

It may also be possible to use `scipy.stats.theilslopes` for trend estimation along with `scipy.stats.kendalltau` for significance testing. Although I'm not sure whether `scipy.stats.kendalltau` is fully equivalent to a Mann-Kendall test. If it is acceptable, then the new option could be `op="theilslopes"` with an output structure such as [`slope`, `intercept`, `low_slope`, `high_slope`, `statistic`, `pvalue`].

### Contribution

- [ ] I would be willing/able to open a Pull Request to contribute this feature.

tlogan2000:
### Pull Request Checklist:
    - This PR fixes #720 

adds new trend analysis operation `theilslopes` to `climatological_op` 

### Other information:
Overall structure might be too messy now?  There are now three sub-functions within climatological_op?

### DIFF
--- a/src/xscen/aggregate.py
+++ b/src/xscen/aggregate.py
@@ -96,8 +96,8 @@ def climatological_op(  # noqa: C901
               ['slope', 'intercept', 'rvalue', 'pvalue', 'stderr', 'intercept_stderr'].
 
             - 'theilslopes' : Computes the Theil-Sen estimator over time, using
-              scipy.stats.theilslopes and employing years as regressors. 
-              Correlation and p-value for the correlation are also computed using scipy.stats.kendalltau, 
+              scipy.stats.theilslopes and employing years as regressors.
+              Correlation and p-value for the correlation are also computed using scipy.stats.kendalltau,
               as the Theil-Sen estimator is based on Kendall's tau.
               The output will have a new dimension 'theilslopes_param' with coordinates:
               ['slope', 'intercept', 'lower_slope', 'upper_slope', 'correlation', 'p_value'].
@@ -167,32 +167,32 @@ def climatological_op(  # noqa: C901
         else:
             out = np.full(6, np.nan)
         return out
-    
-    def _theilslopes(x, y, **kwargs):
-            # Wrapper for scipy.stats.theilslopes to unpack multiple return values in xr.apply_ufunc
-            valid_x = ~np.isnan(x)
-            valid_y = ~np.isnan(y)
-            mask = valid_x & valid_y
-            if np.sum(mask) >= kwargs.get("min_periods", 1):
-                x = x[mask]
-                y = y[mask]
-                reg = scipy.stats.theilslopes(y, x, alpha=kwargs.get("alpha", 0.95), method=kwargs.get("method", "separate"))
-                correlation, p_value = scipy.stats.kendalltau(x, y)
-                out = np.array(
-                    [
-                        reg.slope,  # slope
-                        reg.intercept,  # intercept
-                        reg.low_slope,  # lower bound of slope confidence interval
-                        reg.high_slope,  # upper bound of slope confidence interval
-                        correlation, # correlation coefficient for kendall tau 
-                        p_value, # significance of the correlation for kendall tau
-                    ]
-                )
 
-            else:
-                out = np.full(6, np.nan)
-            return out
-    
+    def _theilslopes(x, y, **kwargs):
+        # Wrapper for scipy.stats.theilslopes to unpack multiple return values in xr.apply_ufunc
+        valid_x = ~np.isnan(x)
+        valid_y = ~np.isnan(y)
+        mask = valid_x & valid_y
+        if np.sum(mask) >= kwargs.get("min_periods", 1):
+            x = x[mask]
+            y = y[mask]
+            reg = scipy.stats.theilslopes(y, x, alpha=kwargs.get("alpha", 0.95), method=kwargs.get("method", "separate"))
+            correlation, p_value = scipy.stats.kendalltau(x, y)
+            out = np.array(
+                [
+                    reg.slope,  # slope
+                    reg.intercept,  # intercept
+                    reg.low_slope,  # lower bound of slope confidence interval
+                    reg.high_slope,  # upper bound of slope confidence interval
+                    correlation,  # correlation coefficient for kendall tau
+                    p_value,  # significance of the correlation for kendall tau
+                ]
+            )
+
+        else:
+            out = np.full(6, np.nan)
+        return out
+
     def _common_trend_utils(ds_rolling=None, func=None, **kwargs):
         # prepare kwargs
         trend_kwargs = {k: v for k, v in op_kwargs.items() if "keep_attrs" not in k}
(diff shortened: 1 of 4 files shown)


tlogan2000:
@aslibese FYI.

tlogan2000:
Just for peace of mind : Quick test on actual data comparing linregress vs theilslopes on homogenized station data:

```
import xarray as xr 
from xscen.aggregate import climatological_op
from xclim import atmos
import matplotlib.pyplot as plt

ds = xr.open_dataset("https://pavics.ouranos.ca/twitcher/ows/proxy/thredds/dodsC/datasets/station_obs/ECCC_AHCCD_gen3_temperature.ncml", chunks={"time": -1, "station": 1})
ds = ds.assign_coords(lat=ds.lat, lon=ds.lon, prov=ds.prov)
f, axs = plt.subplots(1,2, figsize=(10,5))
data = atmos.tg_mean(ds=ds.sel(station=(ds.prov.isin(["QUE"]))))
dsout = xr.Dataset()
dsout[data.name] = data
test = climatological_op(dsout, op="linregress", window=40, periods=['1971','2010'], min_periods=20)
h = axs[0].scatter(x=test.lon, y=test.lat, c=test.tg_mean_clim_linregress.sel(linreg_param='slope'), s=100)
plt.colorbar(h, ax=axs[0])
test2 = climatological_op(dsout, op="theilslopes", window=40, periods=['1971','2010'], min_periods=20)
h2 = axs[1].scatter(x=test2.lon, y=test2.lat, c=test2.tg_mean_clim_theilslopes.sel(theilslopes_param='slope'), s=100)
plt.colorbar(h2, ax=axs[1])
axs[0].set_title("linregress")
axs[1].set_title("theilslopes")
plt.show()

```
image

src/xscen/aggregate.py
```diff
@@ -137,6 +144,105 @@ def climatological_op(  # noqa: C901
...
+
+    def _theilslopes(x, y, **kwargs):
+        # Wrapper for scipy.stats.theilslopes to unpack multiple return values in xr.apply_ufunc
+        valid_x = ~np.isnan(x)
+        valid_y = ~np.isnan(y)
+        mask = valid_x & valid_y
+        if np.sum(mask) >= kwargs.get("min_periods", 1):
+            x = x[mask]
+            y = y[mask]
+            reg = scipy.stats.theilslopes(y, x, alpha=kwargs.get("alpha", 0.95), method=kwargs.get("method", "separate"))
```
juliettelavoie: maybe put in default your are setting here in the docstring

src/xscen/aggregate.py
```diff
@@ -137,6 +144,105 @@ def climatological_op(  # noqa: C901
     If possible, a function that handles NaN values will be used (e.g. op='mean' will use `np.nanmean`), as the
     'min_periods' argument already decides how many NaN values are acceptable.
     """
+
+    def _ulinregress(x, y, **kwargs):
```
juliettelavoie: I think i'd move the subfunctions outside the main function

src/xscen/aggregate.py
```diff
@@ -87,14 +87,21 @@ def climatological_op(  # noqa: C901
...
               The output will have a new dimension 'linreg_param' with coordinates:
               ['slope', 'intercept', 'rvalue', 'pvalue', 'stderr', 'intercept_stderr'].
 
+            - 'theilslopes' : Computes the Theil-Sen estimator over time, using
+              scipy.stats.theilslopes and employing years as regressors.
+              Correlation and p-value for the correlation are also computed using scipy.stats.kendalltau,
+              as the Theil-Sen estimator is based on Kendall's tau.
+              The output will have a new dimension 'theilslopes_param' with coordinates:
+              ['slope', 'intercept', 'lower_slope', 'upper_slope', 'correlation', 'p_value'].
+
```
juliettelavoie, line 107: add a See Also section with link to the external functions (look at xscen.diagnostics.properties_and_measures for an example)
I think you also need to add scipy [here](https://github.com/Ouranosinc/xscen/blob/31d9dbdf3b1082ea3e55e233d7a702f36aa8bea2/docs/conf.py#L120) for it to work

tests/test_aggregate.py
```diff
@@ -651,18 +651,32 @@ def _format(s):
...
     @pytest.mark.parametrize("xrfreq", ["D", "MS", "YS-JAN"])
-    @pytest.mark.parametrize("op", ["max", "mean", "median", "min", "std", "sum", "var", "linregress"])
+    @pytest.mark.parametrize("op", ["max", "mean", "median", "min", "std", "sum", "var", "linregress", "theilslopes"])
     def test_all_default(self, xrfreq, op):
         if op == "linregress" and Version(__scipy_version__) < Version("1.16.0"):
             pytest.skip("Skipping linregress on older scipy")
         o = {"MS": 12, "D": 365, "YS-JAN": 1}[xrfreq]
 
-        ds = timeseries(np.tile(np.arange(1, o + 1), 30), variable="tas", start="2001-01-01", freq=xrfreq, as_dataset=True, calendar="noleap")
+        if op == "theilslopes":
```
juliettelavoie: why do we need a new input here ?

tlogan2000: default dataset used in `linregress` has identical values with slope of zero which returns `nans` for the theilslopes params so created a ds with constant slope of 1 .

src/xscen/aggregate.py
tlogan2000: I've changed this so it is more general and all scipy.stats params can be sent in the `kwargs` for both `linregress` and `theilslopes` options.  For `theilslopes` in particular, as there are two scipy functions being used the `op` parameter needs to explicitly give kwargs for one or both scipy functions eg. `climatological_op(op={"theilslopes": {"theilslopes":{"alpha": 0.90}, "kendalltau": {"method": "auto"}}})`  Have added this info to the doc string

tlogan2000: The default p-value is 0.95 in scipy already so there's no need to really have as before.  scipy call now simply becomes : `scipy.stats.theilslopes(y, x, **kwargs.get("theilslopes", {}))`

src/xscen/aggregate.py
tlogan2000: Done for both linregress and theilslopes (assuming this works as indicated)

src/xscen/aggregate.py
```diff
@@ -871,3 +832,114 @@ def produce_horizon(  # noqa: C901
...
+    return out
+
+
+def _theilslopes(x, y, **kwargs):
+    # Wrapper for scipy.stats.theilslopes to unpack multiple return values in xr.apply_ufunc
+    valid_x = ~np.isnan(x)
+    valid_y = ~np.isnan(y)
+    mask = valid_x & valid_y
+    if np.sum(mask) >= kwargs.get("min_periods", 1):
+        kwargs.pop("min_periods", None)
```
aulemahal:
```suggestion
    if np.sum(mask) >= kwargs.pop("min_periods", 1):
```
nitpick, but FYI you can do that

tests/test_aggregate.py
aulemahal: Maybe we could use this new sloped test data with linregress too ?

aulemahal (approved):
Nice thanks!

tests/test_aggregate.py
tlogan2000: That seems reasonable

tlogan2000: done

PR #733, opened 2026-05-14
Update observational global temperature mean timeseries

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

* Changes how the timeseries are "cleaned" : no more alignment on the 1981-2010 period. Datasets that are absolute (Berkeley) are left as-is. This is mostly esthetic, the anomalies are not modified.
* Adds or modifies the following entries for observational global mean temperature:
    + `Berkeley` becomes `Berkeley-Old`.
    + Remove `Berkeley-raw`, which was useless.
    + New `Berkeley` is the (experimental) updated product from Berkeley Earth that is produced on a 0.25° grid (but we still only download the annual global means)
    + Add `NOAAGlobalTempv6`, the latest updated version from NOAA.
    + Add `Kadow_v100` and `Kadow_v103`. Kadow et al. (2020) describe a method for interpolating missing data using AI and provide timeseries made by running this on HadCRUT.  `Kadow_v100` is the original from the article, based on HadCRUT4 and used in the IPCC AR6 WG1 report. `Kadow_v103` is the latest, based on `HadCRUT5`.
    + Remove `JRA-55`.  Providing this was not useful as the reanalysis doesn't go back to the pre-industrial periods, so any anomalies computed here are actually the ones from Berkeley (which was used to fill the past).

* Adds two meta-realization names for usage in `get_warming_level_from_period` and `get_period_from_warming_level`. 
    + `obs-IPCC-AR6` returns `Berkeley-Old`, `HadCRUT5`, `Kadow_v100` and `NOAAGlobalTempv5`, the same as in the IPCC AR6 WG1 report, chapter 1, figure 1.12. The test verifies we get the exact same warming level for 1986-2005.
    + `obs-IPCC-updated` returns `Berkeley`, `HadCRUT5`, `Kadow_v103` and `NOAAGlobalTempv6`, i.e. the same sources, but with updated versions. I plan to use this in PC.

Yes, because `Berkeley` has changed, while `Berkeley-raw`and `JRA-55` were removed. However, this is small.

### Other information
Using `obs-IPCC-updated` changes the observed warming. Also, the IPCC does a mean, before I did a median.

|   Period  | obs-IPCC-AR6 | obs-IPCC-AR6 | obs-IPCC-updated | obs-IPCC-updated |
|-----------|--------------|--------------|------------------|------------------|
|           | median       | mean         | median           | mean             |
| 1981-2010 | +0.69        | +0.69        | +0.69            | +0.68            |
| 1991-2020 | +0.86        | +0.87        | +0.88            | +0.87            |

FYI @SarahG-579462, first column is what we use currently for PC. Last column is what I would suggest to use in the future.


aulemahal:
Example of usage for the new "meta realization":

```python3
import xscen as xs
import numpy as np

np.mean(xs.get_warming_level_from_period('obs-IPCC-updated', [1991, 2020]))
# np.float64(0.8626058258702498)
```

RondeauG (commented):
Code looks good!
To prevent future confusions:

- `Berkeley-Old` is too vague. If they don't provide version, we could maybe use the NRCan treatment of using the last available year in the data (e.g. `Berkeley-v2025`)?
- Maybe do the same with `Berkeley` ? `Berkeley-Earth` ?
- For Kadow, are `v100` and `v103` taken from somewhere? For the HadCRUT5 version, I seem to see v1 and v2: https://zenodo.org/records/19630991

aulemahal:
Indeed.

Berkeley doesn't provide a version name, but they differentiate the two by their resolution. Maybe : `Berkeley-LowRes` and `Berkeley-HighRes` ?

Oh! I didn't see the Zenodo you linked! I was referring to the Zenodo of the software : https://zenodo.org/records/13767317.  Version 1.0.0 seems to correspond to the article, from which I downloaded the first version and Version 1.0.3b has an updated dataset.

But, I'll use what you found instead and call them `Kadow2020` and `Kadow2026` then.

aulemahal:
Updated and table updated, it seems Kadow2026 is slightly warmer than Kadow v1.0.3b (2024).

RondeauG (approved):

PR #736, opened 2026-05-19
Actually vectorized subsetting on warming levels

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

Rewrites `subset_warminglevel` to use a vectorized `sel` instead of re-building the output with `concats`. Then gain is a drastic decrease in dask complexity.  For a simple case where the input has 15 tasks, I went from 950 to 170. In general, I think the increase in dask tasks is x10 for simple inputs (not much done since opening), compared to x63 before.

This was achieved by building an array to pass to `da.sel(time=...)` instead of doing this part by part. Involves a few `apply_ufunc`, but I hope the code is still readable ?

Also, I added support for non-yearly frequencies here. It's no big deal for a scalar subset, but still requires an uniform time coordinate on the vectorized subset since we need to align all elements on the same "fake" time axis at the end.

Finally, I changed the definition of the `warminglevel_bounds` variable. In CF, common bounds are supposed to be identical for two non-overlapping but contiguous periods. This means that the right (end) bound should not be "1 second before the end", it should simply be the end, so midnight on January 1st of the year following the period. This change has been made in xclim for `time_bnds` a while ago, in both cases it was because of my faulty reading of the conventions.

Should not, except for `warminglevel_bounds`.

### Other information:
@SarahG-579462  Is this useful to you ?  Anyway, it will help PC's crunch a lot.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,7 +3,7 @@ Changelog
 =========
 
 `Unreleased <https://github.com/Ouranosinc/xscen>`_ (latest)
------------------------------------------------------------
+------------------------------------------------------------
 Contributors: Pascal Bourgault (:user:`aulemahal`).
 
 New features and enhancements
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,7 @@ Contributors: Pascal Bourgault (:user:`aulemahal`).
 New features and enhancements
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 * Performance enhancement of ``xs.subset_warminglevels`` when multiple realizations and/or multiple warming levels are requested. The change reduces the dask complexity of the function and adds support for subsetting timeseries with any frequencies. The function still can't do vectorized subsetting on sub-monthly timeseries with non-uniform calendars. (:pull:`736`).
+   + Also added a ``min_periods`` argument to allow subsetting periods shorter than the ``window``.
 
 .. _changes_0.15.0:
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,18 +2,14 @@
 Changelog
 =========
 
-..
-    `Unreleased <https://github.com/Ouranosinc/xscen>`_ (latest)
-    ------------------------------------------------------------
-    Contributors:
+`Unreleased <https://github.com/Ouranosinc/xscen>`_ (latest)
+-----------------------------------------------------------
+Contributors: Pascal Bourgault (:user:`aulemahal`).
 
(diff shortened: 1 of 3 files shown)


src/xscen/extract.py
```diff
@@ -1154,136 +1154,138 @@ def subset_warming_level(
...
+        # Weird, but much easier to use xarray/pandas indexing for selecting years and much cleaner code using apply_ufunc
+        def _bounds_to_sel(time, bnds, not_reached):
+            time = xr.DataArray(time, dims=("time",), coords={"time": time})
+            if not_reached:
+                sl = time.isel(time=slice(0, fake_time.size))
+            else:
+                sl = time.sel(time=slice(*bnds))
+            return sl.values
+
+        # from IPython import embed; embed()
```
RondeauG: Cleanup
```suggestion
```

RondeauG (approved):
LGTM! I don't think that it's too hard to understand with `apply_ufunc` and if it makes it faster, I'm all for it!

CHANGELOG.rst
```diff
@@ -2,18 +2,13 @@
...
-
-    Changes
-    ^^^^^^^
-    * No change.
-
-    Fixes
-    ^^^^^
-    * No change.
+`Unreleased <https://github.com/Ouranosinc/xscen>`_ (latest)
+-----------------------------------------------------------
```
Zeitsperre: Small, but enough to fail the docs:
```suggestion
`Unreleased <https://github.com/Ouranosinc/xscen>`_ (latest)
------------------------------------------------------------
```

aulemahal: This sphinx is so sensitive!

SarahG-579462:
Cool ! Seems similar to my code for PC's backend, but I'm stuck without xscen for now because of zarr3 incompatibility.

How do you handle the case where the rolling average is above a given WL, but there are less than 30 (or N) years in the rolling average? It seems like you reject the entire period with `_wl_not_reached`? I did mine differently by accepting uneven sized subsets (with NaNs where unavailable)

aulemahal:
Indeed, we're lacking a `min_periods` here. I simply preserved the previous behaviour. I can add this!

aulemahal:
@SarahG-579462 @RondeauG  Modified it to add possibility of extracting a period shorter than the window. As I use `sel` to extract the subset, I needed to make some pirouettes to make it work. I use a "sentinel" date that represents invalid timesteps, I chose something that is highly improbable to be found in the dataset. 

This addition increases the dask complexity, but only in the case where `min_periods < window`, otherwise performance is unaltered.

PR #741, opened 2026-05-27
fix docstrings

RondeauG:

* Fixes the RtD warnings in https://github.com/hydrologie/xhydro/pull/432.


Zeitsperre (approved):
I'm convinced this is a sphinx bug. Will look into this.

PR #763, opened 2026-09-01
Add possibility to "unstack" iterable columns in catalog

### LINKED ISSUE #762: Coarser freq with multiple variable rejected in favor of finer freq with single variable
### Setup Information

- xscen version: 0.15.2
- Python version: 3.12
- Operating System: Linux

### Description

In my catalog I have data at hourly and monthly frequencies. The hourly ds are single-variable datasets, while the monthly ds are multi-variable ds (`XYZ`, `XYZ_min`, `XYZ_max`, etc).

When searching for `XYZ` at `MS` with `allow_resampling=True`, the hourly is preferred over the monthly.

### Steps To Reproduce

```python3
cat = xs.DataCatalog('catalog.json')
_, scat = xs.search_data_catalogs(cat, {'mrros': 'MS'}, allow_resampling=True).popitem()
```

### Additional context

[catalog.json](https://github.com/user-attachments/files/31700447/catalog.json)
[catalog.csv](https://github.com/user-attachments/files/31700446/catalog.csv)

In this particular catalog, `mrros` is missing one year (1982) at the hourly step, this is how I realized it wasn't taking the MS files!

### Contribution

aulemahal:
### Pull Request Checklist:
    - This PR fixes #762 

* `DataCatalog.unstack(col)` will take the iterables in "col" and create one entry per element, copying all other fields over.
* `DataCatalog.stack(col)` does the opposite : for all groups where all columns except "col" are the same, merge entries by making "col" a tuple.
* `search_data_catalogs` now will `unstack` before filtering and `stack` (if needed) at the end. 

`search_data_catalogs` has one step where it keeps only the coarsest frequencies for each variable in the result. But by grouping on "variable" is doesn't see that "(XYZ,)" is similar to "(XYZ, ABC)". That was my bug in #762 .

In general, I think this can help for some filtering stuff. Once unstacked, the datasets seems to open correctly, even though the same paths are opened multiple times.  So, re-stacking is not required, one can keep the non-iterable variable column.

I don't think so, but the test suite might not cover all `search_data_catalogs` cases...

### Other information:
The way this is done might not be compatible with the newest intake esm... We'll see!

This also introduces a new sample catalog and dataset to use in the tests. We didn't have any multi-variable catalog.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,7 @@ Contributors:  Juliette Lavoie (:user:`juliettelavoie`).
 New features and enhancements
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 * Add CanESM5-1 to IPCC_annual_global_tas.nc. (:pull:`761`).
+* Add ``DataCatalog.unstack`` and ``DataCatalog.stack`` to transform iterable columns into multiple entries (and the opposite). (:issue:`762`, :pull:`763`).
 
 .. _changes_0.15.2:
 
--- a/src/xscen/catalog.py
+++ b/src/xscen/catalog.py
@@ -621,6 +621,59 @@ class DataCatalog(intake_esm.esm_datastore):
         data = data.drop(columns=["new_path"])
         return self.__class__({"esmcat": self.esmcat.model_dump(), "df": data})
 
+    def unstack(self, col="variable"):
+        """
+        For catalogs where 'col' is an iterable, creates multiple entries for each of the iterable elements.
+
+        Useful for splitting variable columns before filtering. The resulting catalog might not work correctly
+        with dataset creation, in which case :py:meth:`stack` should be called first.
+
+        Parameters
+        ----------
+        col : str
+            Name of the column to "unstack". Raises an error is the column is not iterable.
+        """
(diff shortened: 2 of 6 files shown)


RondeauG (approved):
LGTM! 

I don't remember the reasoning behind having an iterable `variable` column, instead of having multiple rows that point to the same file. Would it be simpler to go back on that?

aulemahal:
The iterable column comes from Intake-esm. I think the idea is that one entry in the catalog corresponds to one asset/file. My tests of opening/extracting datasets did work, but I think that the "unstacked" catalog kinda breaks the ESM catalog idea (multiple entries for a single asset). 

D'ailleurs, it broke one of the notebooks, so might be a sign it's not ideal haha. I'll try to fix this.

aulemahal:
@RondeauG looks like my idea somewhat broke the derived variable handling in `search_data_catalogs`. It seems `intake-esm` only sets `_requested_variables` if the variable column is a list.  Indeed, the need for it does not apply for single-variable assets.

However, it seemed to me that it made more sense to always set these fields ? So I added that to our override of `search`. Now, `_requested_variables`, `_requested_variables_true` and `_dependent_variables` are always set by a search on variables.

RondeauG:
Fair enough, that makes sense!

PR #764, opened 2026-09-03
fix for xclim v1

### LINKED ISSUE #771: clean_up change_attr_prefix doesn't work with ""
### Setup Information

- xscen version: main
- Python version: 3.13
- Operating System: linux

### Description

`clean_up` `change_attr_prefix` doesn't work with "".
If I want to replace cat:xxx for xxx, but it did not work.

 "" is read as False in the if statement. 
I will fix it in the fix-for-xclimv1 PR.

### Steps To Reproduce

```
ds.attrs['cat:source']='model'
xs.clean_up(ds, change_attr_prefix='')
```

### Additional context

I was expecting this to work in ESPO v2 and ESPO-G6-post2100, but it did not work.
So, the outputs have attrs cat:source instead of source. 

PVI @tlogan2000 @huard 

### Contribution

juliettelavoie:
### Pull Request Checklist:
    - This PR fixes #771 

* xclim v1 broke xscen. this fixes it. hopefully.
* Add xs.unstack_ensemble_member
* Fix #771


juliettelavoie:
I think I have fixed most of the issues except for 

1. diagnostics (probably something that needs to be fixed in xsdba)
  a) When loading a yml, properties are not recognized (`ValueError: Can't find requested base class xsdba.properties.quantile`).
  b) just calling a property (xsdba.properties.mean(ds.tas)) fails (in an env with this branch, xclim main and xsdba@xclim_v1)
  `TypeError: StatisticalProperty._preprocess_and_checks() takes 3 positional arguments but 4 were given
`
Do these things work for you @coxipi ?

2. I changed `load_xclim_module` for `load_xclim_collection`. However, now `reload=False` doesn't make much sense. `hasattr(xc.indicators, name)` is always False. @aulemahal  how do we call a collection without reloading it ?

Apart from these issues, all tests, notebook and templates run well.

EDIT:

3. I also have an issue with the wind functions in conversion when running the test extract test_allow_conversion. It looks like when we do a deep copy of the catalog, the var_name that I added are not seen. I am having trouble understanding why.

aulemahal:
2. My idea was that you create the collection and keep it somewhere to use it, you don't rely on monkeypatching to find it later. 

I would argue to remove the  `reload` argument from `load_xclim_collection` and _always_ rebuild the module when this function is called.  En fait, that would make this function useless and it might be cleaner to simply remove it ? 

 However, we could load the conversion one once upon import in xscen and store it somewhere. Maybe  `xscen.indicators.conversion` ?  And then use it for example when `search_data_catalogs` has `allow_conversion=True`.

As a side note, I realized that idea would work better if the indicators in a collection were not registered by default, which is something I will open a PR for in xclim.

src/xscen/indicators.py
```diff
@@ -333,29 +294,33 @@ def select_inds_for_avail_vars(
...
         logger.debug("Loading indicator module.")
-        indicators = load_xclim_module(indicators, reload=True)
+        indicators = xc.IndicatorCollection.from_yaml(indicators)
     if hasattr(indicators, "iter_indicators"):
         indicators = [(name, ind) for name, ind in indicators.iter_indicators()]
+    elif isinstance(indicators, dict):
+        indicators = [(name, ind) for name, ind in indicators.items()]
     elif isinstance(indicators, list | tuple) and not is_list_of_tuples:
-        indicators = [(ind.base, ind) for ind in indicators]
+        # TODO: not really sure what base was ? is identifier the correct new thing to call here ?
```
aulemahal: I'm 50% confident `ind.base` was some kind of mistake ? My error I guess, but it didn't make sense here and was fragile as not defined on all indicators.

src/xscen/indicators.py
```diff
@@ -333,29 +294,33 @@ def select_inds_for_avail_vars(
...
+        indicators = [(ind.identifier, ind) for ind in indicators]
 
+    # TODO: je ne comprends pas le fixme?
     # FIXME: Remove if-else when updating minimum xclim version to 0.53
     XCVARS = xc.core.VARIABLES if hasattr(xc.core, "VARIABLES") else xc.core.utils.VARIABLES
```
aulemahal:
```suggestion
    XCVARS = xc.core.VARIABLES
```
Prior to 0.53, `VARIABLES` was in `xc.core.utils`. It is now in `xc.core` directly. As you pin v1, the test can go.

aulemahal:
Last commit with last commit in the xsdba branch reads measure from the output attributes. However, I tested with nonsensical stuff and it seems the two modified lines are not being tested in `test_diagnostics.py`, FYI.

juliettelavoie:
> Last commit with last commit in the xsdba branch reads measure from the output attributes. However, I tested with nonsensical stuff and it seems the two modified lines are not being tested in `test_diagnostics.py`, FYI.

Yes, it failed previously because it could not find `history` but there is not a test specifically with ratio. I can add something.

juliettelavoie:
With xclim's main and xsdba branch xclim_v1, the tests, templates and notebooks all work.

